PR #3, opened 2015-01-22
Fog of War Prototype

randomPoison:
Added a prototype for the Fog of War. It generates a ring of clouds around a fixed center point, and generates a ring of box colliders to prevent the player from moving outside that area. The fog can be regenerated at runtime, so we can make it larger when you get more lildudes. Fog of war is generated at runtime, but there's a button in the inspector to preview what the FoW will look like.

I also made some tweaks to Chris's original CloudWave shader. Before it would flip the model inside out, so I gave it a min and max value to oscilate between. I've made 4 materials with this shader and gave them slightly different values so that when we generate clouds for the FoW they don't all move in sync.


Assets/Scripts/FogOfWar.cs
```diff
@@ -0,0 +1,138 @@
...
+	public GameObject colliderPrefab;
+
+	public float radius;
+	public float radiusError;
+	public float height;
+
+	[Range( 0.0f, 360.0f )]
+	public float cloudIncrement;
+
+	public float minScale;
```
brickman1444, line 18: Should these be within a range? 0 - 1.0f?

Assets/Scripts/FogOfWar.cs
```diff
@@ -0,0 +1,138 @@
...
+	public float maxScale;
+
+	public int numClouds;
+	public int numColliders;
+
+	private new Transform transform;
```
brickman1444, line 24: what does private _new_ Transform do?

Assets/Scripts/FogOfWar.cs
```diff
@@ -0,0 +1,138 @@
...
+		ClearFogOfWar();
+		GenerateClouds();
+		GenerateColliders();
+	}
+
+	public void ClearFogOfWar( bool inEditor = false )
+	{
+		// destroy all children
+		var children = new List<GameObject>();
+		foreach ( Transform child in transform ) children.Add( child.gameObject );
```
brickman1444, line 47: this is a style

Assets/Scripts/FogOfWar.cs
```diff
@@ -0,0 +1,138 @@
...
+		children.ForEach( child => Destroy( child ) );
```
brickman1444, line 48: is this LINQ?

Assets/Scripts/FogOfWar.cs
randomPoison, line 18: It's an angle in degrees, so 0 - 360.

Assets/Scripts/FogOfWar.cs
randomPoison, line 24: `MonoBehavior` already has a member called `transform`, so the new one I declared hides that member. C# will give a warning when you hide a member, so you use the `new` keyword to tell it you meant to do so.

Assets/Scripts/FogOfWar.cs
randomPoison, line 47: on the `foreach` loop? That's something that I copied from an example. I didn't bother splitting it into multiple lines but if you don't want to do one-line ifs and loops then I'll break it up.

Assets/Scripts/FogOfWar.cs
randomPoison, line 48: [nope](https://msdn.microsoft.com/en-us/library/bwabdf9z%28v=vs.110%29.aspx), just a lambda.

Assets/Scripts/FogOfWar.cs
```diff
@@ -0,0 +1,138 @@
...
+				) as GameObject ).GetComponent<Transform>();
+			float width = 2 * radius * Mathf.Sin( Mathf.Deg2Rad * increment * 0.5f );
+			collider.localScale = new Vector3( width, height, 1.0f );
+			collider.SetParent( transform, false );
+		}
+	}
+
+	void OnDrawGizmos()
+	{
+		Gizmos.DrawIcon(GetComponent<Transform>().position + Vector3.up * .5f, "S.png", true);
```
brickman1444: If we're acknowledging that "transform" is inefficient, then you should be caching it in Awake() unless you're somehow giving the object a different transform at some point.

brickman1444: Create a new icon and name it something relevant.

Assets/Scripts/FogOfWar.cs
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+[CustomEditor( typeof( FogOfWar ) )]
+public class FoWEditor : Editor
+{
+	public override void OnInspectorGUI()
```
brickman1444, line 121: This stuff looks like it's good but I've never done editor scripting before.

PR #4, opened 2015-01-23
Chris dev

cjacobwade:


randomPoison:
Make sure you write a description of your pull request. It should be as comprehensive as reasonable so that it's easier to look over your changes and verify that they make sense.

randomPoison:
Everything here looks solid to me. Thanks for adding the `#if UNITY_EDITOR` stuff to the behavior tree code since that needed to be done. Did you run the game after you merged in master to make sure nothing broke?

Also, why did you comment out all of Avatar.cs? If you want to remove it but are worried about possibly needing it again in the future you can just delete it since you can still recover the old versions through Git.

PR #10, opened 2015-01-28
Simple Buddy Behaviors and Actor Refactoring

randomPoison:
I'm not done with the current round of buddy behavior work, but I want to get the Actor code changes in now.  I decouple the player input code from the main Actor code and everything runs as before. @cjacobwade please look at the changes I've made and let me know what doesn't make sense or should still be changed.

The big changes are:
- Added `PlayerActor` and `PlayerActorPhysics` as subclasses of `Actor` and `ActorPhysics`, respectively. These now handle the player-specific code that was previously being done in `Actor` and `ActorPhysics` (mostly handling input).
- Removed as much of the combat-related code and assets as I could find.
- Added `Actor` and `ActorPhysics` components to the Buddy prefab and changed the behavior tree code to use `ActorPhysics` for movement.
- Removed the behavior tree library code and added it as a submodule so that we can continue to develop in as a separate project.
- Cleaned up some of the existing leaf nodes to make them more generic.


randomPoison:
Take a look at `ActorPhysics.IsGrabbing()` and see if it's used anywhere. If not delete it.

Assets/Scripts/Actors/PlayerActor.cs
```diff
@@ -0,0 +1,20 @@
+﻿using UnityEngine;
+using System.Collections;
+
+[RequireComponent(typeof(ActorCamera))]
+public class PlayerActor : Actor
+{
+	ActorCamera _actorCamera;
```
randomPoison, line 7: This should also have `ActorResources`.

PR #11, opened 2015-01-29
Formatting Configuration

randomPoison:
@brickman1444 @cjacobwade I've setup a Clang-Format config file for our current code style and formatted ActorPhysics.cs and PlayerActorPhysics.cs. Have a look over the way they're formatted and let's discuss a finalized code style.

As for using clang-format as a formatting tool, it doesn't fully support C# so it doesn't handle `public`,`protected`,`private` correctly (it treats them like c++ and puts them on a separate line). If you only highlight the body of a function and format it then it works fine, though. You guys don't have to use it, but I think we should include the .clang-format file to document our style standards.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,35 @@
+IndentWidth: 4
+TabWidth: 4
+UseTab: ForIndentation
+
+NamespaceIndentation: All
+AccessModifierOffset: 0
+
+ColumnLimit: 0
+
+BreakBeforeBraces: Allman
+# BreakBeforeBinaryOperators: All
+BreakBeforeTernaryOperators: false
+
+AlignTrailingComments: true
+AlignOperands: true
+
+SpaceInEmptyParentheses: false
+SpacesInCStyleCastParentheses: false
+SpaceAfterCStyleCast: false
+SpacesInContainerLiterals: false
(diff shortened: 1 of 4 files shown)


cjacobwade:
How does this affect workflow? I want to keep things agile and not worry about line spacing and running a formatting tool every commit...

randomPoison:
We're going to be working on this codebase for a while (probably at least a year), and if our code is messy and inconsistent it's going to start slowing things down over time. Things like line spacing may not seem important, but they affect readability, and inconsistent style increases the mental load of trying to understand a piece of code. Consistent organization within files makes it easier to find something in a file you didn't write. You don't have to run clang-format if you don't want to, it's easy enough to format your code as you type it. It's just a nice thing to have since you don't have to worry about the details as much.

cjacobwade:
I feel like you're over emphasizing the importance of formatting in game code. There's a limit to how much it matters and stuff like if( variable ) vs if(variable) and space instead of tabs seem to be too granular. If clang does that stuff for me and it's a one or two button thing, that's fine but I really can't imagine myself scraping code for spacing and line ending differences.

cjacobwade:
Curious to hear what Zac has to say on this also.

PR #13, opened 2015-01-30
Buddy Emotes

randomPoison:
- Buddies now emit a heart emote when given a resource.


brickman1444:
Looks good to me.

randomPoison:
@cjacobwade Adding a `RequireComponent( Animator )` attribute to the `BuddyStats` script won't work because the animator is supposed to be on one of the child objects, not on the root buddy object. Plus the animator still needs to be setup from the inspector with the correct animation, so we may as well get clear errors if the object is setup incorrectly.

PR #16, opened 2015-01-31
Buddy Death

randomPoison:
- Pressing 'K' kills buddies.
- Once dead, buddies to not move and will not accept resources.


Assets/Scripts/Actors/BuddyStats.cs
```diff
@@ -13,10 +19,21 @@ void Awake()
...
+		{
+			Kill();
+		}
+	}
+
 	public void GiveResource( ActorPhysics actorPhysics, ResourceData resourceData )
 	{
-		currentStats++;
-		Emote( heartMaterial );
+		if ( _isAlive )
```
brickman1444: Put a comment or maybe an console error message saying that we should revisit this and put an assert in here.

Assets/Scripts/Actors/BuddyStats.cs
```diff
@@ -25,4 +42,11 @@ public void Emote( Material emoteMaterial )
 		_particles.renderer.material = emoteMaterial;
 		_particles.Emit( 1 );
 	}
+
+	public void Kill()
+	{
+		_isAlive = false;
+		Destroy( GetComponent<AIController>() );
+		GetComponentInChildren<Animator>().SetTrigger( "isDead" );
```
brickman1444, line 58: Maybe leave a comment explaining what happens after you do this stuff.

brickman1444:
That's everything.

PR #17, opened 2015-02-02
Collecting Buddies and Working Inventory

randomPoison:
- I refactored the inventory system to allow for non-resource items, specifically so that you could pick up the buddy eggs.
- Added a buddy egg item which can be picked up in the world and used to create a new buddy.
- Added an assert utility and a math utility for doing correct modulo calculations.


cjacobwade:
I think InventoryItem should hold InventoryItemData instead of ResourceData so the player can carry the egg. Somewhere this will probably also need to fit pet or hug options from the player.

cjacobwade:
Do you think putting the menu button methods into the data class they're creating would make more sense? Upside is deleting that class removes the method, downside is you migth need to add "if editor" tags...

cjacobwade:
Looks good otherwise

randomPoison:
Looks like I already changed InventoryItem. As for the menu buttons, I'd like to avoid using `#if EDITOR` blocks as much as possibly, since Unity wraps up the functionality in the Editor folder and separating them keeps the code more readable.

PR #20, opened 2015-02-03
Stress test

brickman1444:
Here's the setup for the stress test. I don't know if we actually want this in the repo or not.


randomPoison:
I'm fine with having a scene for doing stress tests as long as it's separate from the main scene.

cjacobwade:
Same here, go for it.

PR #21, opened 2015-02-03
Singleton Update

brickman1444:
Update to the singleton that we're using with the changes from the other repo. Not very pressing. Add any other thoughts you guys have on this.


randomPoison:
I think the persistent singleton functionality should be split off into a separate class because it can be simplified by automatically doing `DontDestroyElseKill()` in `Awake()`. Beyond that, I'm not sure when we would need a non-persistent, MonoBehaviour singleton, but it won't hurt to keep the functionality around.

brickman1444:
I'm going to leave this as is. We can split it the persistent singleton off if we start using singletons more and decide that it'd be better.

PR #23, opened 2015-02-04
Remove UniMerge

randomPoison:
- Remove .gitattributes and .gitconfig files and the merge-unity scripts that were causing the merge problems.
- Remove UniMerge from the .gitignore so that people are notified if they have it locally.

As it turns out, the UniMerge integration with Git doesn't work. So even if UniMerge works perfectly, we cant properly invoke it when we need to without going through a bunch of trouble on our end, so we should just drop it. We can go back to the old-fashioned method of merging scenes/assets, and we can mitigate conflicts by taking some additional measures within Unity.


randomPoison:
@cjacobwade @brickman1444 we'll have to talk about workflow and how to best handle dealing with unity assets until I have a better solution.

brickman1444:
Looks good.

PR #24, opened 2015-02-04
Sound manager edits

cjacobwade:
Use prefab for getting sound data with soundmanager calls, added mesh combine utility for batching later, audio zone prefab that starts playing on player enter and stops on player exit (for efficient ambient noise)


Assets/Scripts/Actors/ActorComponents/PlayerActorResources.cs
```diff
@@ -174,7 +174,7 @@ void OnTriggerEnter( Collider other )
 			PickupResource( resourceComponent.resourceData );
 
 			Destroy( other.gameObject );
-			WadeUtils.TempInstantiate( resourcePopPrefab, other.transform.position, Quaternion.identity, 1f );
+			SoundManager.instance.Play3DSoundAtPosition( resourcePopAudio, other.transform.position );
```
randomPoison: @brickman1444 How do we want to do singletons in the project? I'd argue that we have public static members that automatically act on the instance, rather than having to manually retrieve the instance every time you use the singleton. Chris and I discussed this at GGJ, but I don't think we've brought it up for this project.

Assets/Scripts/Actors/FootstepSound.cs
```diff
@@ -1,9 +1,13 @@
...
+public class FootstepSound : MonoBehaviour 
+{
+	[SerializeField] AudioSource[] footstepSounds;
 
-	[SerializeField] void PlayerWalkSound(){
-		SoundManager.instance.PlaySoundAtPosition ("Footstep", transform.position-transform.up);
+	[SerializeField] void PlayerWalkSound()
+	{
+		SoundManager.instance.Play3DSoundAtPosition ( footstepSounds[Random.Range(0, footstepSounds.Length)],
+		                                              transform.position - transform.up );
```
randomPoison: Why are you doing `transform.position - transform.up`? What's gained by the extra offset?

Assets/Scripts/Utilities/AudioZone.cs
```diff
@@ -0,0 +1,31 @@
...
+
+	void OnTriggerEnter( Collider col )
+	{
+		if ( col.GetComponent<PlayerActor>() )
+		{
+			playingClip = SoundManager.instance.Play3DSoundAndFollow( sourceData, transform );
+		}
+	}
+
+	void OnTriggerExit( Collider col )
```
randomPoison: Please use a better variable name than `col`. It's not much harder to use a longer name since we have auto-complete, and it's good practice to use descriptive variable names.

Assets/Scripts/Utilities/AudioZone.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections;
+
+public class AudioZone : MonoBehaviour 
+{
+	[SerializeField] AudioSource sourceData = null;
+	AudioSource playingClip = null;
+
+	void OnTriggerEnter( Collider col )
+	{
+		if ( col.GetComponent<PlayerActor>() )
```
randomPoison: Rather than looking for `PlayerActor` specifically, we can make this more reusable by creating a tag component that indicates the AudioZone should activate.

cjacobwade:
This is my preferred way of doing it and I'd argue it makes the singleton code way more readable. Since that's been the driving factor in defining our code standard I think we should keep it this way.

Assets/Scripts/Actors/ActorComponents/PlayerActorResources.cs
cjacobwade: This is my preferred way of doing it and I'd argue it makes the singleton code way more readable. Since that's been the driving factor in defining our code standard I think we should keep it this way.

Assets/Scripts/Utilities/AudioZone.cs
```diff
@@ -0,0 +1,31 @@
...
+		{
+			playingClip = SoundManager.instance.Play3DSoundAndFollow( sourceData, transform );
+		}
+	}
+
+	void OnTriggerExit( Collider col )
+	{
+		if ( col.GetComponent<PlayerActor>() && playingClip.gameObject )
```
randomPoison: Why check `playingClip.gameObject` for null and not just `playingClip`? Is there ever a situation where a component would exist but its game object wouldn't?

Assets/Scripts/Actors/FootstepSound.cs
cjacobwade: This plays the footstep sound on the players feet rather than their mid point.

Assets/Scripts/Utilities/MeshCombine.cs
```diff
@@ -0,0 +1,26 @@
...
+[RequireComponent(typeof(MeshRenderer))]
+public class MeshCombine : MonoBehaviour 
+{
+	void Start() 
+	{
+		MeshFilter[] meshFilters = GetComponentsInChildren<MeshFilter>();
+		CombineInstance[] combine = new CombineInstance[meshFilters.Length];
+
+		int i = 0;
+		while ( i < meshFilters.Length )
```
randomPoison: A for-loop would be preferable here since it would be more idiomatic for the iteration you're doing.

Assets/Scripts/Utilities/AudioZone.cs
cjacobwade: I use col rather than collider because monobehaviour uses collider already. I guess I can use triggerCol, but col seems descriptive enough to me in the context of an OnTrigger function.

Assets/Scripts/Utilities/SoundManager.cs
```diff
@@ -2,254 +2,127 @@
...
 
-	public void SetupPool(GameObject soundManagerObj)
+	public void SetupPool( GameObject soundManagerObj )
 	{
 		// Create object to hold sounders
-		audioObjHolder = new GameObject(poolName);
+		audioObjHolder = new GameObject( "AudioPool" );
 		audioObjHolder.transform.parent = soundManagerObj.transform;
+		
+		loopObjHolder = new GameObject( string.Concat( "AudioPool", "_Loops" ) );
```
randomPoison: Why are you doing `string.Concat()` and two string literals? Also, using the `+` operator is preferable to `string.Concat()` because it is more readable and they compile to the same code: http://www.dotnetperls.com/string-concat

randomPoison:
I'd like to take this as an opportunity to revise the sound manager and see if the suggestions I had previously made are applicable. Please don't merge this in until I've had a chance to do so.

Assets/Scripts/Utilities/SoundManager.cs
```diff
@@ -2,254 +2,127 @@
...
 
-    // Use this for initialization
-    void Awake()
+	public AudioSource Play3DSoundAtPosition( AudioSource sourceData, Vector3 position )
     {
-		foreach(AudioPool audioPool in audioPools)
+		foreach ( AudioSource audioSource in audioObjs )
 		{
-			foreach(AudioClip audioClip in audioPool.audioClips)
+			Debug.Log( audioSource );
```
randomPoison: Remove this. Only generally useful logging information should be put on master.

Assets/Scripts/Utilities/SoundManager.cs
```diff
@@ -2,254 +2,127 @@
...
-				audioPool.audioObjs.Add(newSource);
-				return PlayAudioObj(newSource, clip);
-			}
-		}
+		AudioSource newSource = CreateAudioObj();
+		newSource.transform.position = position;
+		audioObjs.Add( newSource );
 
-		return null;
+		return PlayAudioObj( sourceData, newSource );
```
randomPoison: I'd like this function to be refactored so that it only has one `return` in it. Instead of returning form the loop, use the loop to find the first available obj. If there is none, create one, then set its position:

```
AudioSource source;

foreach ()
{
    // set source if an unused audio source is found
}

if ( !source )
{
    // create new audio source object and set source to reference it.
}

// set values and return source
```

This keeps the logical flow of the function clean and cuts down on code duplication. It's not super important for this particular function, but it's a good practice to maintain.

Assets/Scripts/Utilities/SoundManager.cs
```diff
@@ -2,254 +2,127 @@
...
     }
 
-	public AudioSource PlaySoundAndFollow(string clip, Transform target)
+	public AudioSource Play3DSoundAndFollow( AudioSource sourceData, Transform target )
     {
-		foreach(AudioPool audioPool in audioPools)
+		foreach ( AudioSource audioSource in audioObjs )
```
randomPoison: this foreach-loop to find the first available audio source is used in multiple places, so it should be moved into a helper method.

cjacobwade:
Which suggestions do you mean?

Assets/Scripts/Utilities/SoundManager.cs
```diff
@@ -2,254 +2,127 @@
...
-		yield return new WaitForSeconds(audioSource.clip.length);
+		yield return new WaitForSeconds( audioSource.clip.length );
 
-		ResetSource(audioSource);
-        audioSource.transform.position = parent.position;
-		audioSource.transform.parent = parent;
+		if( !audioSource.loop )
+		{
+			ResetSource( audioSource );
+	        audioSource.transform.position = parent.position;
```
randomPoison: This line is indented with spaces (should be tabs).

Assets/Scripts/Utilities/WadeUtils.cs
```diff
@@ -402,6 +404,36 @@ public static GameObject TempInstantiate(GameObject obj, Vector3 pos, Quaternion
 		return go;
 	}
 
+	public static T GetCopyOf<T>(this Component comp, T other) where T : Component
```
randomPoison, line 435: Be careful about using reflection in utility code, especially since this is in a class with a bunch of other utility functionality. Reflection isn't supported on all platforms (mostly mobile has problems with it), so when you start using it you risk causing problems when building for release. Probably not important for this game, just something to be aware of.

Assets/Scripts/Utilities/AudioZone.cs
cjacobwade: fixed

Assets/Scripts/Actors/FootstepSound.cs
randomPoison: We should probably have the transform root be at the players feet then. If the player is standing at (0, 0, 0), they should be standing on top of the x-z plane, not chest-deep in the middle of it.

Assets/Scripts/Utilities/AudioZone.cs
randomPoison: Hmm, that's a fair point. Something like `otherCol` or even just `other` might be better, though.

randomPoison:
Using `Instantiate()` rather than reflection to copy the values, and using the Unity Gems object pool rather than a home-rolled implementation.

Assets/Scripts/Utilities/AudioZone.cs
cjacobwade: I can this up with a bitfield that indicates which types this zone accepts and then then the zone tag includes an enum field. That way certain zones can accept only gods, only buddies, only the player, etc. Although maybe that enum field could just be on actor. Let me know when you're on chat so we can discuss this a bit.

Assets/Scripts/Utilities/AudioZone.cs
cjacobwade: fixed

Assets/Scripts/Utilities/MeshCombine.cs
cjacobwade: fixed

Assets/Scripts/Utilities/SoundManager.cs
cjacobwade: fixed

Assets/Scripts/Utilities/AudioZone.cs
randomPoison: Yeah, we can talk about how best to do that. At the least I'd like to not tie it down to this particular game so that we can reuse the sound manager in the future without having to make any modifications.

Assets/Scripts/Utilities/SoundManager.cs
cjacobwade: fixed

cjacobwade:
Like I said, instantiating a component makes a copy of the gameObject. http://stackoverflow.com/questions/19506133/why-instantiate-rigid-body-instead-of-gameobject-in-unity

Assets/Scripts/Utilities/WadeUtils.cs
cjacobwade, line 435: Interesting. Do you have a source on mobile having problems w/ it? I'd like to read more about it.

Assets/Scripts/Utilities/SoundManager.cs
cjacobwade: Refactored this a lot. Made a helper that does this instead, but good call. Way cleaner.

Assets/Scripts/Actors/FootstepSound.cs
cjacobwade: Makes sense to me

Assets/Scripts/Utilities/SoundManager.cs
cjacobwade: fixed

Assets/Scripts/Utilities/SoundManager.cs
cjacobwade: fixed

Assets/Scripts/Actors/ActorComponents/PlayerActorResources.cs
brickman1444: I vote for wrapping up the public singleton interfaces with static methods that call private non static methods. e.g. static void Singleton::foo() { instance.privFoo(); }

That makes it easy to read the calling code and the implementation code. You lose the knowledge on the calling end that you're using a singleton but that shouldn't really matter.

randomPoison:
btw don't worry about changing the player prefab to be at the right height, I'll take care of that as part of reworking the movement code.

Assets/Scripts/Actors/ActorComponents/PlayerActorResources.cs
cjacobwade: I like that. Good way to get best of both. I'll go through and add that.

cjacobwade:
I figured. I don't plan on touching anything physics related until you've had a crack at it. Let me know when you look into your changes, so we can get this in.

randomPoison:
Aaaaactually, I think the physics overhaul is going to take up too much time to be able to address this in a reasonable amount of time. I'm going to open up an issue to remind me to work on the sound manager. Otherwise you're good to merge this in if you've addressed all the comments.

cjacobwade:
Sounds like a plan.

PR #26, opened 2015-02-10
Rossen physics

brickman1444:
Changing the actor to use the Rossen Physics with the lifter and bumper. Should be reasonably usable at the moment.


Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -12,28 +12,32 @@ public enum ActorStates
 
 public class ActorPhysics : ActorComponent
 {
+	private new Transform transform;
```
cjacobwade, line 16: Why do this instead of naming a new variable? I get that we don't want to be doing getComponent all the time, but this seems hacky.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -43,52 +47,47 @@ public float groundedMoveSpeed
...
-	float groundSlopeCapsuleWidth = 0.4f;
-	float groundSlopeCastDist = 0.3f;
-
-	float groundSlopeFallSpeed = -30f;
+	[Range( 0.0f, 1.0f )]
+	[SerializeField] float lookDirLerp = 0.002f;
 
-	protected float groundSlopeSpeedMod = 1f;
-	float groundSlopeSpeedModeDelta = 2.5f;
+	private Vector3 _lastPos;
```
cjacobwade, line 75: Why are we labeling private variable with underscores? Looks weird and doesn't really give any extra info. If a variable is private, I'll notice when I try to use it.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -301,40 +346,6 @@ public void StopClimbing()
 		}
 	}
 
-	protected void CheckGroundSlope()
```
cjacobwade, line 304: Removing this brings back the problem of sliding up Unity terrain. It's still happens and I think we need a solution before we remove this.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -362,15 +373,19 @@ public void JumpCheck()
...
+		                 -Vector3.up,
+		                 out hit,
+		                 jumpCheckDistance,
+		                 jumpLayer );
+		if ( hit.transform && Vector3.Dot( hit.normal, Vector3.up ) > minJumpDot )
 		{
 			isOnGround = true;
 		}
 
-		if ( ( isOnGround || lateJumpTimer < lateJumpTime ) )
```
cjacobwade, line 371: Bring back late jump before committing please

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -380,77 +395,92 @@ public void JumpCheck()
...
-			{
-				if ( !IsInState( ActorStates.Grounded ) )
-				{
-					// if !launching
-					ChangeState( ActorStates.Grounded );
-					stopMoveTimer = 0f;
-					//GainControl();
-				}
-
-				actor.animator.SetBool( "isJumping", false );
```
cjacobwade, line 393: This is removing the cue for the jump animation so this will need to be added back in.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -380,77 +395,92 @@ public void JumpCheck()
...
-	public void ModelControl()
+	/**
+	 * @brief Orients the model to face the direction of movement
+	 * 
+	 * @details
+	 *     Orientation is reactive to velocity, meaning the
+	 *     model will face the direction of movement, rather
+	 *     than the input direction.
+	 */
+	public void OrientSelf()
```
cjacobwade: Something in here or in climbing causes a weird bug where the player model rotation is offset and doesn't return to normal.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -380,77 +395,92 @@ public void JumpCheck()
...
-		Gizmos.DrawSphere( transform.position - Vector3.up * groundSlopeCheckHeight + model.forward * 0.3f + model.right * 0.2f, groundSlopeCheckRadius );
-		Gizmos.DrawSphere( transform.position - Vector3.up * groundSlopeCheckHeight + model.forward * 0.3f - model.right * 0.2f, groundSlopeCheckRadius );
-
-		// Climb check
-		Gizmos.color = Color.green;
-		Gizmos.DrawSphere( transform.position, climbCheckRadius );
-		Gizmos.DrawSphere( transform.position + model.forward * climbCheckDistance, climbCheckRadius );
+		if ( Application.isPlaying )
+		{
+			Gizmos.DrawWireSphere( GetComponent<Transform>().position + bumper.center,
```
cjacobwade: Why should this only be when playing? This goes back to why we shouldn't use new private Transform. Calling GetComponent<Transform> looks totally weird.

Assets/Scripts/Tags/ClimbableTag.cs
```diff
@@ -3,16 +3,16 @@
 
 public class ClimbableTag : MonoBehaviour 
 {
-	[SerializeField] bool xMovement;
-	public bool XMovement
+	[SerializeField] bool _xMovement = true;
```
cjacobwade: I like this use case where the private variables accessed through an instance have underscores, but don't think usual private variables need an underscore.

Assets/Scripts/Actors/ActorComponents/PlayerActorPhysics.cs
```diff
@@ -34,7 +32,11 @@ public override void SetupStateMethodMap()
 
 	void Jumping()
 	{
-		JumpCheck();
+		// TODO should we be doing JumpCheck() while jumping?
```
cjacobwade: GetButtons during FixedUpdate sometimes miss button presses. Should change this to set a flag every frame if input is down.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
brickman1444, line 16: The idea is that the interface is the same but with only one GetComponent() call.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
brickman1444, line 75: Yea David and I have tried to start doing it. It helps when writing a system to quickly visually see what's public and private.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade, line 16: It's just weird to me that it depends on Awake for transform to work. If anything else tries to refer to this before awake it will get a null ref. I'd prefer something separate like myTransform so the getComponent functionality is still there as well.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
brickman1444, line 393: I'm not totally sure what's going on here but it looks like it's a case where it is exiting the jump state and reentering grounded. So it's not triggering the jump animation, it's stopping it. Does it still seem necessary?

cjacobwade, line 393: There's a big logic chunk that got cut off. We basically need to put logic back in to control animation states somewhere since it's taken out here.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
brickman1444: changed to if ( transform )

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 16: Yeah, I've noticed that when doing things like gizmos and editor scripting it causes problems. Unity also has `Reset()` which get's called in the editor, but if that doesn't work for some reason we could do lazy initialization like we're doing `instance` for `Singleton`.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 304: The lifter/bumper combo should take care of this. If the slope is steep enough Then it will collide with the bumper and push the player back from the wall. We can tweak the maximum climbable slope by changing the configuration of the colliders.

randomPoison:
@cjacobwade @brickman1444 I just want to make it clear that the physics refactoring is **not** done yet, that is why there are so many problems and things that have not been implemented.

cjacobwade:
Yep! That was clear, I didn't look to much at syntax. Mostly just marking TODOs.

randomPoison:
Ah, cool.

randomPoison:
Also, btw, it looks here like a bunch of important things were deleted, but they weren't. I had them commented out while working on other stuff and Zac deleted them in his cleanup commit. I'll be rolling back his changes and putting back most of that code.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: Yeah, I've got a fix. I changed `ActorPhysics` to rotate the root transform rather than the model's transform, but there were other places in code that were still rotation the model. I cleaned those up and that fixed the problem.

randomPoison:
I made a bunch more changes. Jump delay has been added back in and various issues have been fixed. The climbing system still is in need of a cleanup pass in light of the new collisions, but most stuff is working now and platforming feels pretty good. @brickman1444 @cjacobwade Give this another look over if you have a chance before we create the build, otherwise it's ready for the show build at least.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade, line 16: How would the implementation for lazy initialized this work? Wouldn't you need another variable anyway?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -6,34 +6,39 @@ public enum ActorStates
...
 public class ActorPhysics : ActorComponent
 {
+	private new Transform transform;
+	private new Rigidbody rigidbody;
+
 	// States
+	#region States
 	public delegate void ActorStateMethod();
 
+	[ReadOnlyAttribute, SerializeField]
```
cjacobwade: Can we name attributes without the word Attribute in them? It would be preferable not to since it can be assumed, but I might recall that being needed...

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -6,34 +6,39 @@ public enum ActorStates
...
+	#region Movement
+	[Header( "Movement" )]
 	public Vector3 inputVec = Vector3.zero;
 	public float moveSpeedMod = 1f;
 
-	protected Vector3 lastVelocity = Vector3.zero;
 	protected Vector3 moveVec = Vector3.zero;
 
-	protected float stoppingSpeed = 0.001f;
+	public float stoppingSpeed = 0.001f;
```
cjacobwade: This shouldn't be public. It's referenced in ActorPlayerPhysics but nowhere else.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -43,113 +48,135 @@ public float groundedMoveSpeed
...
-	protected Vector3 modelOffset;
-
-	// Slope Checking
-	[Space(10)][Header("Slope Checking")]
-	[SerializeField] float groundSlopeCheckRadius = 0.2f;
-	[SerializeField] float groundSlopeCheckHeight = 0.7f;
-	float groundSlopeCapsuleWidth = 0.4f;
-	float groundSlopeCastDist = 0.3f;
+	[Range( 0.0f, 1.0f )]
+	[SerializeField] float lookDirLerp = 0.002f;
```
cjacobwade: Can you rename this to something readable in the inspector?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: Yeah, that should actually be just `ReadOnly`, I'm surprised having `ReadOnlyAttribute` didn't cause a build error.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -6,34 +6,39 @@ public enum ActorStates
...
 
-	protected float stoppingSpeed = 0.001f;
+	public float stoppingSpeed = 0.001f;
 	protected float currStoppingPower = 0.0f;
 
 	[SerializeField] float stopMoveTime = 0.3f;
 	float stopMoveTimer = 0f;
 
-	[SerializeField] float maxSpeed = 40f;
 	[SerializeField] float _groundedMoveSpeed = 6f;
```
cjacobwade, line 41: Either all or non of these private, serialized vars should have the underscore.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -43,113 +48,135 @@ public float groundedMoveSpeed
...
 
-	[SerializeField] float jumpCheckDistance = 1.3f;
-	[SerializeField] Vector3 groundPosOffset = new Vector3(0f, -1f, 0f);
-	
-	[SerializeField] float jumpCheckRadius = 1.2f;
+	[SerializeField] float jumpCheckDistance = 1.0f;
+	[SerializeField] float jumpCheckRadius = 0.5f;
+
 	[SerializeField] LayerMask jumpLayer = 0;
 	[SerializeField] float minJumpDot = 0.4f;
```
cjacobwade: This could probably have a better name. It's clear if dot product is second nature, but totally obtuse otherwise.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -43,113 +48,135 @@ public float groundedMoveSpeed
...
-	// Climbing
-	[Space(10)][Header("Climbing")]
+	#region Climbing
+	[Space( 10 ), Header( "Climbing" )]
 	[SerializeField] LayerMask climbLayer = (LayerMask)0;
-	[SerializeField] float climbCheckDistance = 0.5f;
 	[SerializeField] float climbCheckRadius = 0.7f;
-	[SerializeField] float surfaceHoldForce = 0.1f;
 
 	[SerializeField] float climbCheckTime = 0.2f;
```
cjacobwade: This isn't used anymore in your implementation.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -43,113 +48,135 @@ public float groundedMoveSpeed
...
 	float climbCheckTimer = 1f;
```
cjacobwade: climbCheckTime/climbCheckTimer aren't used anymore with your implementation.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -43,113 +48,135 @@ public float groundedMoveSpeed
...
 
 	Transform climbSurface = null;
-
-	public bool isGrabbing
-	{
-		get{ return WadeUtils.ValidAxisInput("Grab"); }
-	}
+	ClimbableTag climbTag;
+	Vector3 climbSurfaceNormal = Vector3.zero;
+	Vector3 climbSurfaceRight = Vector3.zero;
```
cjacobwade: If we have climbSurface saved can't we just get these when we need them? Would be easy to write your z-in/out-of-wall check to be reusable.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -6,34 +6,39 @@ public enum ActorStates
...
+	[Header( "Movement" )]
 	public Vector3 inputVec = Vector3.zero;
 	public float moveSpeedMod = 1f;
 
-	protected Vector3 lastVelocity = Vector3.zero;
 	protected Vector3 moveVec = Vector3.zero;
 
-	protected float stoppingSpeed = 0.001f;
+	public float stoppingSpeed = 0.001f;
 	protected float currStoppingPower = 0.0f;
```
cjacobwade: This is no longer being used anywhere.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -182,101 +209,99 @@ public void ComeToStop()
...
-				surfaceRelativeInput = new Vector3( surfaceRelativeInput.x, 
-				                                    surfaceRelativeInput.z, 
-				                                    surfaceRelativeInput.y );
+			currStoppingPower = stoppingSpeed;
 
-				if ( !ct.XMovement ) surfaceRelativeInput.x = 0f;
-				if ( !ct.YMovement ) surfaceRelativeInput.y = 0f;
+			Vector3 surfaceRelativeInput =
+			    climbSurfaceRight * ( climbTag.xMovement ? movement.x : 0.0f ) +
+			    Vector3.up * ( climbTag.yMovement ? movement.z : 0.0f );
```
cjacobwade: Formatting is weird here, probably should split this up.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -182,101 +209,99 @@ public void ComeToStop()
...
 
-				model.rotation = Quaternion.Lerp( model.rotation, 
-				                                  Quaternion.LookRotation( climbSurface.forward, Vector3.up ), 
-				                                  Time.deltaTime * modelTurnSpeed );
-			}
-			else
-			{
-				Debug.LogError("Cannot climb, surface isn't tagged. This is probably a problem in ClimbCheck");
-			}
+			transform.Translate( moveVec * Time.deltaTime, Space.World );
```
cjacobwade: This makes me so unhappy, but only because I hate transform.Translate irrationally. I'd totally set velocity here.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -182,101 +209,99 @@ public void ComeToStop()
...
+			Debug.LogWarning( "Climb volume is facing into wall." );
+		}
+
 		ChangeState( ActorStates.Climbing );
+		
+		rigidbody.useGravity = false;
+		rigidbody.velocity = Vector3.zero;
+		lifter.gameObject.SetActive( false );
+		bumper.gameObject.SetActive( false );
+		climbBumper.gameObject.SetActive( true );
```
cjacobwade, line 315: Probably should split these into a function like SetClimbingCollidersActive.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -286,13 +311,16 @@ public void StartClimbing( Collider col )
...
+		if ( climbSurface )
 		{
-			rigidbody.useGravity = true;
 			climbSurface = null;
-			ChangeState( ActorStates.Jumping );
+			climbTag = null;
+			rigidbody.useGravity = true;
+			lifter.gameObject.SetActive( true );
+			bumper.gameObject.SetActive( true );
+			climbBumper.gameObject.SetActive( false );
```
cjacobwade, line 332: Then use SetClimbingCollidersActive(false) here.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade, line 304: Did you get this feeling right or are we going with invisible walls on every steep slope? We should mention it to Tasha if so.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -403,54 +409,113 @@ public void JumpCheck()
...
+	 * model will face the direction of movement, rather
+	 * than the input direction.
+	 */
+	void OrientSelf()
+	{
+		Quaternion desiredLook = transform.rotation;
+		if ( IsInState( ActorStates.Climbing ) )
 		{
-			jumpColCheckTimer += Time.deltaTime;
+			Quaternion.LookRotation( climbSurfaceNormal );
```
cjacobwade: Guessing this should be desiredLook = Quaternion.LookRotation( climbSurfaceNormal );

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -403,54 +409,113 @@ public void JumpCheck()
...
 		}
+		else
+		{
+			Vector3 actualVelocity = transform.position - _lastPos;
+			actualVelocity.y = 0.0f;
+
+			Vector3 intendedVelocity = rigidbody.velocity;
+			intendedVelocity.y = 0.0f;
+
+			Vector3 lookVec = Vector3.Lerp( actualVelocity, intendedVelocity, lookDirLerp );
```
cjacobwade: lookDirLerp should probably be named something like lookIntentionWeight.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -403,54 +409,113 @@ public void JumpCheck()
...
+		EndLateJump();
+	}
 
-	public void ModelControl()
+	void EndLateJump()
 	{
-		model.position = transform.position - modelOffset;
+		_lateJump = false;
+	}
+	#endregion
```
cjacobwade, line 539: Hadn't thought of using Invoke this way. Clever! We should wrap this in a Timer class and have like StartTimer, EndTimer, CheckTimer and CancelTimer with an enum as a timerID.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -403,54 +409,113 @@ public void JumpCheck()
...
-		// SlopeCheck
-		Gizmos.color = Color.blue;
-		Gizmos.DrawSphere( transform.position - Vector3.up * groundSlopeCheckHeight + model.forward * 0.3f + model.right * 0.2f, groundSlopeCheckRadius );
-		Gizmos.DrawSphere( transform.position - Vector3.up * groundSlopeCheckHeight + model.forward * 0.3f - model.right * 0.2f, groundSlopeCheckRadius );
-
-		// Climb check
-		Gizmos.color = Color.green;
-		Gizmos.DrawSphere( transform.position, climbCheckRadius );
-		Gizmos.DrawSphere( transform.position + model.forward * climbCheckDistance, climbCheckRadius );
+		if ( Application.isPlaying )
```
cjacobwade: Rip this out before merging.

Assets/Scripts/Actors/ActorComponents/PlayerActorPhysics.cs
```diff
@@ -96,30 +117,33 @@ void JumpMovement()
...
-
-			moveVec = inputVec * jumpMoveSpeed * groundSlopeSpeedMod;
+			moveVec = inputVec * jumpMoveSpeed;
 			moveVec.y = rigidbody.velocity.y;
 
-			lastVelocity = moveVec;
 			rigidbody.velocity = moveVec;
 
-			if ( actor.animator )
+			if ( _actor.animator )
```
cjacobwade: Other spots have _actor.animator != null, I prefer this shorter way of doing it. Either way pick one and be consistent.

Assets/Scripts/Utilities/DebugUtils.cs
```diff
@@ -8,7 +8,7 @@ public static void Assert( bool condition, string message = "" )
 		if ( !condition )
 		{
 			Debug.LogError( message );
-			System.Diagnostics.Debugger.Break();
+			// Debug.Break(); // Uncomment if you want to pause the editor when an assert is hit.
 		}
```
cjacobwade, line 12: This function should have a bool like BreakOnAssert with default value false. Then Debug.Break only happens if that's true.

cjacobwade:
There's a bunch of commented code all over and the underscoring is somewhat inconsistent. I say lets do them on all private variables serialized or not.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: They're just commented out, there's big performance hit when not using them but it feels way worse when using the timers, so I want to leave this in until I can put together a better solution.

PR #28, opened 2015-02-10
Multiple Player Stats

brickman1444:
Genericates the player stats so that we can easily add new stats and have a common interface for changing them. Updates the system so the stamina is updated through the new system. Buddies are now tagged with the stat they update.

Adds GrossUtils.cs which includes Serializeable Dictionary and Serializeable HashSets.


Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -188,7 +188,7 @@ public void ComeToStop()
 
 	public void ClimbCheck()
 	{
-		if (isGrabbing && ( ( actor as PlayerActor ) == null || ( actor as PlayerActor ).actorStats.CanUseStamina() ) )
+		if (isGrabbing && ( ( actor as PlayerActor ) == null || ( actor as PlayerActor ).actorStats.CanUseStat( Stat.Stamina ) ) )
```
cjacobwade, line 191: Consider caching actor to a PlayerActor var

randomPoison, line 191: Yeah, this is something that I'm taking care of in the physic refactoring. Checking stats shouldn't even be happening within `ActorPhysics`, it should take place in `PlayerActorPhysics`.

randomPoison:
GrossUtils.cs has a bunch of places where spaces aren't padded. Also let's stop this whole MyNameUtils thing.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
brickman1444, line 191: Yea I'll let the physics refactoring take care of this.

brickman1444:
Fixed formatting issues in GrossUtils. Made issue #35 for consolidating utilities.

PR #29, opened 2015-02-10
Tech art

cjacobwade:
Adding color utilities, adding render settings adjustment test script, render settings manager and render settings volumes, adding shader for variable color and brightness.


Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -0,0 +1,137 @@
+﻿using UnityEngine;
+using System.Collections;
+
+[System.Serializable]
```
randomPoison: Do `using System;` instead of manually doing `System.Serializable` every time.

randomPoison:
Fix the formatting issues and you're good.

PR #36, opened 2015-02-11
Adding Gitmerge

randomPoison:
- Added the GitMerge plugin as a submodule so that we can use it for development but I can still work on improving it.


brickman1444:
Looks good.

PR #37, opened 2015-02-11
Tech art

cjacobwade:
Adding skybox shifting functionality, skybox shifting zones and environment shaders


Assets/Scripts/ImageEffects/EagleVision.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections;
+
+[ExecuteInEditMode]
+public class EagleVision : MonoBehaviour 
+{
+	[SerializeField] Material mat;
+
+	void OnRenderImage (RenderTexture source, RenderTexture destination)
+	{	
+		if(mat)
```
brickman1444: pad with spaces

Assets/Scripts/ImageEffects/EagleVision.cs
```diff
@@ -0,0 +1,61 @@
...
+		{
+			Graphics.Blit(source, destination, mat, 0);
```
brickman1444: spaces here too

Assets/Scripts/ImageEffects/EagleVision.cs
```diff
@@ -0,0 +1,61 @@
+﻿using UnityEngine;
+using System.Collections;
+
+[ExecuteInEditMode]
+public class EagleVision : MonoBehaviour 
+{
+	[SerializeField] Material mat;
```
brickman1444: start names of private variables with underscores

Assets/Scripts/ImageEffects/EagleVision.cs
```diff
@@ -0,0 +1,61 @@
...
+
+	void OnRenderImage (RenderTexture source, RenderTexture destination)
+	{	
+		if(mat)
+		{
+			Graphics.Blit(source, destination, mat, 0);
+		}
+
+//		if(CheckResources() == false) {
```
brickman1444, line 17: Why's all this commented out? Is it a work in progress?

cjacobwade:
Sorry, I just started the eagle vision stuff so that's not ready yet.

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -3,29 +3,25 @@
...
-[Serializable]
-public class FogSettings
-{
-	public Color color;
-	public float density;
+	public Color lightColor = Color.white;
+	public float lightIntensity = 0.5f;
+
+	public Color fogColor = Color.white;
+	public float fogDensity = 0.05f;
```
brickman1444, line 14: Are these default values chosen on purpose or are they left from prototyping?

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -3,29 +3,25 @@
 using System;
 
 [Serializable]
-public class LightSettings
+public class TimeRenderSettings
```
brickman1444, line 6: Did you consider making this a struct?

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -36,7 +32,9 @@ public class RenderSettingsManager : SingletonBehaviour<RenderSettingsManager>
 
 	Light _dirLight = null;
 
-	[Range(0, 1)]
+	[SerializeField] float dayCycleTime = 200;
```
brickman1444: Again. Are these proper default values or just left over from something?

brickman1444: Also is this seconds? Ticks? Fortnights?

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -46,6 +44,15 @@ void Awake()
 
 	void Update()
 	{
+		dayCycleTimer += Time.deltaTime;
+
+		timeOfDay = (-Mathf.Cos(dayCycleTimer/dayCycleTime * Mathf.Rad2Deg) * 0.5f) + 0.5f;
```
brickman1444: what's the Rad2Deg doing in there? Mathf.Cos takes in radians. http://docs.unity3d.com/ScriptReference/Mathf.Cos.html

brickman1444: Also leave a comment specifying the range of possible values for timeOfDay. It takes a bit to figure out if you're looking at it for the first time.

brickman1444: and add padding

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -64,18 +71,28 @@ void IChangeTargetRenderSettings( RenderSettingsData renderSettings )
 
 	void ApplyRenderSettings()
 	{
-		Color skyboxColor = Color.Lerp( _currentRenderSettings.dayColor,
-		                                _currentRenderSettings.nightColor,
+		Color skyboxColor = Color.Lerp( _currentRenderSettings.daySettings.color,
+		                                _currentRenderSettings.nightSettings.color,
 		                                timeOfDay );
 		RenderSettings.skybox.SetColor( "_Tint", skyboxColor );
```
brickman1444: is that the only way to change stuff on a skybox? Strings are gross

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -64,18 +71,28 @@ void IChangeTargetRenderSettings( RenderSettingsData renderSettings )
...
 
-		RenderSettings.fogColor = _currentRenderSettings.fogSettings.color;
-		RenderSettings.fogDensity = _currentRenderSettings.fogSettings.density;
+		RenderSettings.fogColor = Color.Lerp( _currentRenderSettings.daySettings.fogColor,
```
brickman1444, line 99: I'm not seeing where this "RenderSettings" comes from. Is it a class with static fields?

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -86,34 +103,53 @@ void ApplyRenderSettings()
 	void LerpRenderSettings( RenderSettingsData initRenderSettings, float delta )
```
brickman1444, line 123: Is this function still used? If so, where? If not, do we still need it?

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade, line 6: Will change.

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade, line 14: These were chosen on purpose. I think 0.5f is the default light setting in Unity, 0.05f is an obvious fog and then White is an obvious color. Do you think I should use something that looks like a real setting as the default or something obscene so we actually set these?

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade: I was having trouble getting values I expected so I went with something that looked right. I'll go back over this tonight.

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade: Leftover, will fix.

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade: That's how you change the settings on a material. There's a way to do it with something called nameID. I'll have to figure that out at some point.

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade, line 99: It's a built in unity class with static fields

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade, line 123: It's used in the coroutine at the end of that script. This is used by render volumes

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444, line 14: I'm cool with what you have. I was just curious if these are actual usable default values. I'd rather have "totally usable defaults" or "these are insane" instead of "these defaults are kind of okay". All these seem to fall into the former category so you're good.

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: Ok. From my understanding it would make sense to put 2Pi there. Maybe it's a scaling issue with the day length? ie the day length as is doesn't really correspond to a real value like seconds because it is being scaled by that deg2rad.

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: Ok

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444, line 99: Ah okay.

Assets/Scripts/ImageEffects/EagleVision.cs
cjacobwade: Fixed.

Assets/Scripts/ImageEffects/EagleVision.cs
cjacobwade: Fixed.

Assets/Scripts/ImageEffects/EagleVision.cs
cjacobwade: Fixed.

Assets/Scripts/ImageEffects/EagleVision.cs
cjacobwade, line 17: Added comment explaining in progress-ness.

brickman1444:
Should be good to go once you finish up those fixes.

Assets/Scripts/Utilities/RenderSettingsManager.cs
cjacobwade: Ended up needing to multiply by 2pi, not sure why that wasn't obvious when I was doing it the first time...

PR #38, opened 2015-02-12
Cutting

brickman1444:
Three cutting prototypes in a separate scene.
Not very well done code organization. I know it.


cjacobwade:
Cutting is sweet. I like the cut and then regrow after a while best. I think that one will feel really good once we have a path filled with them. Can you have the grass check if the player is on it before regrowing?

Assets/Scripts/CuttingPrototypes/CutableB.cs
```diff
@@ -0,0 +1,31 @@
...
+
+	public void Cut( float cuttingLevel )
+	{
+		if ( health / cuttingLevel <= magicRatioNumber )
+		{
+			health -= cuttingLevel;
+
+			GameObject particleObj = (GameObject) Instantiate( _particlePrefab,
+															  transform.position + new Vector3( 0, _verticalOffset, 0 ),
+															  Quaternion.identity );
```
cjacobwade: Some weird formatting here.

Assets/Scripts/CuttingPrototypes/CutableA.cs
```diff
@@ -0,0 +1,23 @@
...
+	[SerializeField] GameObject _particlePrefab = null;
+	[SerializeField] float _particleLifetime = 0;
+	[SerializeField] float _verticalOffset = 0;
+
+	public void Cut( float cuttingLevel )
+	{
+		if ( cuttingLevel >= _level )
+		{
+			Destroy( gameObject );
+			GameObject particleObj = (GameObject)Instantiate( _particlePrefab,
```
cjacobwade, line 17: Is no spaces in casting operator canon?

Assets/Scripts/CuttingPrototypes/CuttingA.cs
```diff
@@ -0,0 +1,89 @@
...
+						{
+							Cut( cutableComponentC );
+						}
+						else
+						{
+							Debug.LogError( "Attach Cutable component to " + cutableObj.name + " at " + cutableObj.transform.position );
+						}
+					}
+				}
+			}
```
cjacobwade, line 71: I know this code is going to get changed, but a better way to organize this would be to have the CutablesA,B,C inherit from a base Cubtable and just call cut directly w/ polymorphism.

Assets/Scripts/CuttingPrototypes/CutableA.cs
brickman1444, line 17: Yea that's the way I've been doing it. I like how it reads.

Assets/Scripts/Utilities/GrossUtils.cs
```diff
@@ -65,27 +65,15 @@ public void OnAfterDeserialize()
...
 		{
-			if ( values.Count == 0 )
-			{
-				values.Add( new TValue() );
-			}
-			else
-			{
-				values.Add( values[values.Count - 1] ); // duplicate the value key
-			}
+			keys[keys.Count - 1] = new TKey();// If the keys list is extended in the inspector, it will copy the last key in the list
```
cjacobwade, line 74: What's going on here? This looks like you'll end up with two copies of the same key.

Assets/Scripts/CuttingPrototypes/CutableB.cs
brickman1444: fixed. it was a tabs vs spaces issue.

Assets/Scripts/CuttingPrototypes/CuttingA.cs
brickman1444, line 71: Yea I thought the same thing, but we'll have no use for that sort of structure after next Tuesday when the final one is in and this was faster to prototype with.

Assets/Scripts/Utilities/GrossUtils.cs
brickman1444, line 74: The problem was that Unity automatically duplicates the last item in a list when you increase the size. So if you increased the size of the keys list it copies the last key, and THEN you have duplicate keys. So I'm catching that case here and setting that last item in the list to a new key that is the result of the default constructor.

Should all that be made clearer in the code?

brickman1444:
I left out the check for if the player is there because with that current set up there is no way to make a bush truly block the player. The player could always run through it right as it was cut down. Then even if it had the shortest possible dead-time it wouldn't respawn until the player was through it. So even the hardest bushes couldn't block the player.

Assets/Scripts/Utilities/GrossUtils.cs
cjacobwade, line 74: That makes sense now that you've explained it. Maybe make the comment more idiot proof?

Assets/Scripts/CuttingPrototypes/CutableA.cs
cjacobwade, line 17: I'm cool with that

cjacobwade:
So you're saying the difficulty comes from the respawn timer? I feel like that's going to suck as a player if they chop through it and then it immediately regrows and they can't pass by. Maybe we can have visible partial growth when you're on it and it goes back into place once you're on it?

cjacobwade:
This is good to go for our build tomorrow so go ahead and ship it when you're ready.

brickman1444:
Yea for that variation, the difficulty comes from the respawn timer. I think the most satisfying final version will take a little of each of these.

PR #41, opened 2015-02-13
Adds ExportOBJ Script

randomPoison:
- Adds option to file menu to export selected GameObject or prefab as an OBJ mesh.
- Also you can right click on the Transform component in the inspector to export the object.

### DIFF
--- a/Assets/Editor/ExportOBJ.cs
+++ b/Assets/Editor/ExportOBJ.cs
@@ -8,19 +8,13 @@ public class ObjExporterScript
 {
 	private static int StartIndex = 0;
 
-	public static void Start()
-	{
-		StartIndex = 0;
-	}
-	public static void End()
+	public static void Reset()
 	{
 		StartIndex = 0;
 	}
 
 	public static string MeshToString( MeshFilter mf, Transform t )
 	{
-		Vector3 s = t.localScale;
-		Vector3 p = t.localPosition;
 		Quaternion r = t.localRotation;
 
 		int numVertices = 0;
@@ -73,18 +67,11 @@ public class ObjExporterScript
 
 public class ObjExporter : ScriptableObject
 {
-	[MenuItem( "File/Export/Wavefront OBJ" )]
+	[MenuItem( "Export/Wavefront OBJ" )]
 	[MenuItem( "CONTEXT/Transform/Wavefront OBJ" )]
 	static void DoExportWSubmeshes()
 	{
 		DoExport( true );
-	}
-
-	[MenuItem( "File/Export/Wavefront OBJ (No Submeshes)" )]
-	[MenuItem( "CONTEXT/Transform/Wavefront OBJ (No Submeshes)" )]
-	static void DoExportWOSubmeshes()
-	{
-		DoExport( false );
 	}
 
 	static void DoExport( bool makeSubmeshes )
@@ -98,7 +85,7 @@ public class ObjExporter : ScriptableObject
 		string meshName = Selection.gameObjects[0].name;
 		string fileName = EditorUtility.SaveFilePanel( "Export .obj file", "", meshName, "obj" );
 
-		ObjExporterScript.Start();
+		ObjExporterScript.Reset();
 
 		StringBuilder meshString = new StringBuilder();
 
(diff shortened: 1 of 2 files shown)


Assets/Editor/ExportOBJ.cs
```diff
@@ -0,0 +1,158 @@
...
+		{
+			meshString.Append( processTransform( t.GetChild( i ), makeSubmeshes ) );
+		}
+
+		return meshString.ToString();
+	}
+
+	static void WriteToFile( string s, string filename )
+	{
+		using( StreamWriter sw = new StreamWriter( filename ) )
```
brickman1444, line 140: What is going on here?

brickman1444:
Besides the one point where it uses C# I don't understand, everything looks fine. Open it up in Visual Studio and check for warnings.

Assets/Editor/ExportOBJ.cs
```diff
@@ -0,0 +1,158 @@
...
+				                          triangles[i + 2] + 1 + StartIndex ) );
+			}
+		}
+
+		StartIndex += numVertices;
+		return sb.ToString();
+	}
+}
+
+public class ObjExporter : ScriptableObject
```
brickman1444, line 68: Why is this a scriptable object? Is that the only way to add menu stuff?

randomPoison, line 68: No, you can add menu items from any class as long as it's a static void method. I think scriptable object is needed to create files and write to them.

Assets/Editor/ExportOBJ.cs
```diff
@@ -0,0 +1,158 @@
...
+		}
+		Material[] mats = mf.renderer.sharedMaterials;
+
+		StringBuilder sb = new StringBuilder();
+
+		foreach ( Vector3 vv in m.vertices )
+		{
+			Vector3 v = t.TransformPoint( vv );
+			numVertices++;
+			sb.Append( string.Format( "v {0} {1} {2}\n", v.x, v.y, -v.z ) );
```
cjacobwade, line 34: David, didn't you say it was faster to use += when concatenating strings? Obviously this code isn't really performant heavy but string.Append is used a bunch in here and it doesn't match our standards.

Assets/Editor/ExportOBJ.cs
cjacobwade, line 140: Looks like it's putting the string s into a file. Using gets an object, makes it readonly for the duration of the block and then disposes of the reference: https://msdn.microsoft.com/en-us/library/yh598w02.aspx

Assets/Editor/ExportOBJ.cs
```diff
@@ -0,0 +1,158 @@
...
+
+		WriteToFile( meshString.ToString(), fileName );
+
+		t.position = originalPosition;
+
+		ObjExporterScript.End();
+		Debug.Log( "Exported Mesh: " + fileName );
+	}
+
+	static string processTransform( Transform t, bool makeSubmeshes )
```
cjacobwade: Should be ProcessTransform

Assets/Editor/ExportOBJ.cs
```diff
@@ -0,0 +1,158 @@
...
+	private static int StartIndex = 0;
+
+	public static void Start()
+	{
+		StartIndex = 0;
+	}
+	public static void End()
+	{
+		StartIndex = 0;
+	}
```
cjacobwade, line 14: Start and end are dumb method names to show the same thing happening...

Assets/Editor/ExportOBJ.cs
brickman1444, line 34: This is using a StringBuilder. I'm not sure if that even has the += operator.

randomPoison, line 34: [This article](http://www.dotnetperls.com/stringbuilder) talks about the performance of string builders down towards the bottom. For the most part, if you're going to do a lot of appending, then `StringBuilder.Append()` is going to be faster than +=. We can still speed this up, though, by telling the string builder from the beginning how big the final string is going to be. Since we know how many vertices and faces there are, we should be able to do that calculation before actually building the string.

Issue #40, opened 2015-02-12
Make PlayerActorPhysics Not Be A Child of ActorPhysics


randomPoison:
After my latest changes the only part of `PlayerActorPhysics` that actually derives from `ActorPhysics` is `SetupStateMethodMap()` which is a virtual call, and that can be easily fixed. `PlayerActorPhysics` should be something like "PlayerControls", then we can wrap up all control/input functionality into one place rather than having it scattered over a bunch of places.

randomPoison:
Woops, this was done forever ago.

PR #44, opened 2015-02-14
Convert PlayerActorPhysics into PlayerControls

randomPoison:
work on #40
- Updated the Player prefab to reflect the changes, and patched some references that got broken at some point.
- Also updated the buddy animator to clear up some warnings.


Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -122,6 +123,12 @@ public float rollMoveSpeed
 	Vector3 _climbSurfaceUp = Vector3.zero;
 	#endregion
 
+	#region Gliding
+	[Space( 10 ), Header( "Gliding" )]
+	[SerializeField] float _glideHorizontalSpeed = 5.0f;
+	[SerializeField] float _glideDescentRate = 1.0f;
```
brickman1444, line 132: Double check these default values. They should either be completely usable defaults or completely unusable. No use in making defaults that could be seen as okay and aren't.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+
+	public void GlideMovement( Vector3 inputVec )
+	{
+		_moveVec = inputVec * _glideHorizontalSpeed;
+		_moveVec.y = -_glideDescentRate;
+
+		rigidbody.velocity = _moveVec;
+	}
+
+	public void RegisterStateMethod( ActorStates state, ActorStateMethod method )
```
brickman1444: Is this the best place in the file for this method? Closer to the top might make more sense

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+		_bumperRigidbody.velocity = rigidbody.velocity;
+	}
+
+	void FollowBumper()
+	{
+		if ( !IsInState( ActorStates.Climbing ) )
+		{
+			Vector3 constrainedPos = transform.position;
+			constrainedPos.x = _bumperTransform.position.x;
+			constrainedPos.z = _bumperTransform.position.z;
```
brickman1444: you could flip this around and do 

```
constrainedPos = bumpertransform.position;
constrainedPos.y = transform.position.y;
transform.position = constrainedpos;
```

Saves you one assignment. The current way might be more readable though.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+			transform.position = constrainedPos;
+		}
+	}
+
+	void MoveAtSpeed( Vector3 inVec, float moveSpeed )
```
brickman1444: Is invec normalized? also what even is invec?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+	{
+		_inputVec = inVec;
+		rigidbody.useGravity = true;
```
brickman1444: Is this necesary to do every time you move?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+
+		_moveVec = inVec * moveSpeed * _moveSpeedMod;
```
brickman1444: what's movespeedmod for?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+		rigidbody.velocity = _moveVec;
+
+		if ( actor.animator )
+		{
+			actor.animator.SetBool( "isMoving", true );
+		}
+	}
+
+	/**
+	 * Provide resonable default values for the state methods.
```
brickman1444: typo on reasonable

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+	}
+
+	void ChangeState( ActorStates toState )
 	{
 		_currentState = toState;
-		CurrentStateMethod = stateMethodMap[_currentState];
+		CurrentStateMethod = _stateMethodMap[_currentState];
+	}
+
+	void StartClimbing( Collider col )
```
brickman1444: is col the climb volume?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -400,10 +347,139 @@ public void JumpMovement( Vector3 inputVec )
...
+		}
+
+		if ( Vector3.Dot( Vector3.up, _climbSurface.up ) > 0.0f )
+		{
+			_climbSurfaceUp = _climbSurface.up;
+		}
+		else
+		{
+			_climbSurfaceUp = -_climbSurface.up;
+			Debug.LogWarning( "Climb volume is upside down." );
```
brickman1444: Are these log warnings just for level design errors? Can these checks be wrapped up in #Editor flags?

brickman1444: Also I'd make these errors if something's wrong every time it happens. I use warnings for when something's probably wrong when it occurs.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -460,7 +539,7 @@ void OrientSelf()
 		if ( IsInState( ActorStates.Climbing ) )
 		{
 			Vector3 lookVector = _climbSurfaceNormal;
-			//lookVector.y *= _leanTowardsSurface;
+			lookVector.y *= _leanTowardsSurface;
```
brickman1444, line 551: Double check for warnings. I was getting an unused variable warning last night. This looked like it fixed it but double check for any others.

Assets/Scripts/Actors/ActorComponents/PlayerActorPhysics.cs
```diff
@@ -1,135 +0,0 @@
...
-	{
-		base.Awake();
-
-		_actor = GetComponent<PlayerActor>();
-		ChangeState( ActorStates.Grounded );
-	}
-
-	void Update()
-	{
-		if ( Input.GetKeyDown( KeyCode.Escape ) )
```
brickman1444, line 21: This could be split into it's own component or combined with the reset scene component. It needs to be somewhere or else a release build won't close when you hit esc.

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+	{
+		_actor = GetComponent<PlayerActor>();
+		_actorPhysics = GetComponent<ActorPhysics>();
+
+		SetupStateMethodMap();
+	}
+
+	void Update()
+	{
+		if ( Input.GetKeyDown( KeyCode.Escape ) )
```
brickman1444, line 23: this is an ok place to put it

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+		{
+			Application.Quit();
+		}
+
+		_isGrabbing = WadeUtils.ValidAxisInput("Grab");
+		_jumpButtonDown = Input.GetButtonDown( "Jump" + WadeUtils.platformName );
+		_glideButtonDown = Input.GetKey( KeyCode.F );
```
brickman1444, line 30: Why are each of these different styles of input getting?

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	void SetupStateMethodMap()
+	{
+		_actorPhysics.RegisterStateMethod( ActorStates.Jumping, Jumping );
+		_actorPhysics.RegisterStateMethod( ActorStates.Falling, Jumping );
+		_actorPhysics.RegisterStateMethod( ActorStates.Grounded, Grounded );
+		_actorPhysics.RegisterStateMethod( ActorStates.Rolling, Rolling );
+		_actorPhysics.RegisterStateMethod( ActorStates.Climbing, Climbing );
```
brickman1444, line 39: consider aligning all the second parameters with spaces

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -23,12 +24,12 @@ public class ActorPhysics : ActorComponent
...
 	ActorStates _currentState = ActorStates.Jumping;
 	ActorStateMethod CurrentStateMethod;
-	public Dictionary<ActorStates, ActorStateMethod> stateMethodMap = new Dictionary<ActorStates, ActorStateMethod>();
+	Dictionary<ActorStates, ActorStateMethod> _stateMethodMap = new Dictionary<ActorStates, ActorStateMethod>();
 	#endregion
 
 	#region Movement
 	[Header( "Movement" )]
-	public Vector3 inputVec = Vector3.zero;
+	public Vector3 _inputVec = Vector3.zero;
```
cjacobwade: This shouldn't be underscored if it's public. You may want to use readonly here?

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+	void Rolling()
+	{
+		_actorPhysics.RollCheck();
+		_actorPhysics.RollMovement();
+	}
+
+	void Climbing()
+	{
+		if ( _actorPhysics.ClimbCheck() &&
+			_isGrabbing &&
```
brickman1444: consider realigning conditions with spaces

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -134,7 +141,12 @@ public override void Awake()
 		_bumperTransform = bumper.GetComponent<Transform>();
 		_bumperRigidbody = bumper.GetComponent<Rigidbody>();
 
-		SetupStateMethodMap();
+		InitializeStateMethodMap();
+	}
+
+	void Start()
+	{
+		ChangeState( ActorStates.Grounded );
```
cjacobwade, line 152: Why do this in start rather than in awake after initialization?

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		_actorPhysics.RollCheck();
+
+		GroundMovement();
+	}
+
+	void Gliding()
+	{
+		if ( _jumpButtonDown )
```
brickman1444, line 108: what happens if you jump while gliding?

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		if ( _isGrabbing )
+		{
+			_actorPhysics.ClimbCheck();
+		}
+
+		if ( !_glideButtonDown )
+		{
+			_actorPhysics.EndGlide();
```
brickman1444: consider removing the not and reversing order of the if and else cases. To me it makes sense to do the positive case first

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade: Yes. Maybe change this to climbCol for even more readability.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade: He chose this way so we'll know it's wrong and should be fixed but it isn't game breaking priority like an error would be.

Assets/Scripts/PlayerControls.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Collections;
+
+[RequireComponent( typeof( ActorPhysics ) )]
+public class PlayerControls : MonoBehaviour
+{
+	PlayerActor _actor;
+	ActorPhysics _actorPhysics;
+	bool _jumpButtonDown = false;
+	bool _glideButtonDown = false;
+	bool _isGrabbing = false;
```
cjacobwade: This might make more sense as _grabbingInputActive. The current name doesn't seem to fit with _jumpButtonDown and _glideButtonDown.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: Yeah, that makes sense.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: It's the direction that you want to move in. I'll rename it to something like `moveDir` since that's more descriptive. As for it being normalized, it shouldn't be because in most cases it's the input value which can scale based on how far the stick is moved over.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: Apparently not, removing it.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: Nothing, apparently. Removing it.

Assets/Scripts/PlayerControls.cs
brickman1444: Agreed. _isGrabbing sounds like a statement about the state of the physics.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: I'm actually removing these because it doesn't really have an impact on gameplay (since we can figure out what the right direction is). If we're really worried about things being placed weirdly in the level, we can write an editor script to check that all the climb volumes at least have their Y axis facing up, since that could actually cause problems if they were facing to the left or something.

Assets/Scripts/Actors/ActorComponents/PlayerActorPhysics.cs
randomPoison, line 21: Yeah, once we have some kind of level manager or something I'd say that's the place to put it. Or if we have something handling pausing and whatnot.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: `readonly` means that the value can only be set in the constructor, so it's not really useful in the context of MonoBehavoiurs since you don't define a constructor. Really that value should be private, `ActorPhysics` shouldn't be exposing data members, only methods.

randomPoison: Yeah, `_inputVec` wasn't really being used, so I'm pulling it out.

Assets/Scripts/PlayerControls.cs
randomPoison: Done.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 152: The way it's setup, the expectation is that another there will be another script that acts as the controller and it will setup the state method map in `Awake()`. `ChangeState()` also sets a reference to the current method, so if it is called before the controller script has had a chance to setup the state methods then it will have a reference to the default method, which does nothing. `InitializeStateMethodMap()` just fills the method map will empty methods to prevent null references.

Assets/Scripts/PlayerControls.cs
randomPoison, line 108: Nothing, unless you're _reaaaaaalllllly_ close to the ground. This is to match the behavior when jumping, which let's you jump again when you're just about on the ground but haven't quite landed yet.

Assets/Scripts/PlayerControls.cs
randomPoison: Yeah, I agree.

Assets/Scripts/PlayerControls.cs
randomPoison: Good point, changing it.

PR #45, opened 2015-02-15
Resource spawning fix

brickman1444:
Fixed resources spawning over edges. Attempted to fix resources spawning inside things but didn't have good results. Seemed like raycasts didn't always collide with the terrain object I was testing on.

### DIFF
--- a/Assets/Scripts/Resources/ResourceSpawner.cs
+++ b/Assets/Scripts/Resources/ResourceSpawner.cs
@@ -11,30 +11,35 @@ public class ResourceSpawner : MonoBehaviour
 	public int total;
 	public float radius;
 
+	private new Transform transform = null;
+
 	void Start()
 	{
 		// configure resource holder for the current resource type
 		resourceHolderPrefab.resource = resourcePrefab;
 
+		transform = GetComponent<Transform>();
+
 		// spawn resource holders
 		for ( int i = 0; i < total; i++ )
 		{
-			Vector3 offset = Quaternion.Euler( -90.0f, 0.0f, 0.0f ) * (Vector3)Random.insideUnitCircle;
-			Vector3 position = transform.position + offset * radius;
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Resources/ResourceSpawner.cs
```diff
@@ -11,30 +11,76 @@ public class ResourceSpawner : MonoBehaviour
...
+	private new Transform transform = null;
+
 	void Start()
 	{
 		// configure resource holder for the current resource type
 		resourceHolderPrefab.resource = resourcePrefab;
 
+		transform = GetComponent<Transform>();
+
+		if ( !resourcePrefab.GetComponent<Collider>() )
```
randomPoison: Have the `resourcePrefab` member be of type `Collider`, then you won't have to do this check.

Assets/Scripts/Resources/ResourceSpawner.cs
```diff
@@ -11,30 +11,76 @@ public class ResourceSpawner : MonoBehaviour
...
+				Vector3 reallyHighUpSpot = spawnPosition + Vector3.up * WadeUtils.LARGENUMBER;
+
+				// Do a raycast from really high up and exclude the renderzone layer
+				hit = WadeUtils.RaycastAndGetInfo( new Ray( reallyHighUpSpot, Vector3.down ), ~(1 << 13), WadeUtils.LARGENUMBER );
+
+				if ( hit.collider && hit.point.y > spawnPosition.y )
+				{
+					// This is supposed to check if the spawn point is within another object but doesn't entirely work
+					continue;
+				}*/
```
randomPoison: Make sure you remove commented out code before merging into master.

randomPoison:
I think the best solution for getting resources to spawn in a good spot is to simply not place the spawner in such a way that the resources will be floating in the air or whatever. We should add a visualization of the radius in which the resources are being spawned (via `OnDrawGizmos()`) and then just instruct our level designers to not place spawners in such a way that they'll be in a bad spot. Then we only need to do one raycast down to settle them onto the ground.

PR #46, opened 2015-02-15
Combine cutting prototypes into solidified mechanic

brickman1444:
Combines all the cutting prototypes into one solidified mechanic.
Adds particle tinting and a different death particle
Adds sound effects for hits and deaths.

Try out the scene and see how it feels.


cjacobwade:
Is there a specific scene I should be looking at for this? I get errors in the main scene where playerPhysics is turned off, no colliders are assigned for bumpers, lifters or climb bumper. Also the bushes give errors for missing particle assignment when you slash them.

Assets/Scripts/Cutting/Cutting.cs
```diff
@@ -0,0 +1,61 @@
...
+				Cuttable cuttableComponent = cuttableObj.GetComponent<Cuttable>();
+
+				if ( cuttableComponent )
+				{
+					Cut( cuttableComponent );
+				}
+
+				else
+				{
+					Debug.LogError( "Attach Cuttable component to " + cuttableObj.name + " at " + cuttableObj.transform.position );
```
randomPoison, line 51: We should probably just write an editor script that checks if any objects are on the `cuttable` layer and don't have a `Cuttable` component on them.

PR #49, opened 2015-02-16
Build Fix

randomPoison:
- The Player prefab didn't have the Gliding stat yet, so you'd get a crash when trying to glide.
- Also I forgot to register the Gliding() method with ActorPhysics.


cjacobwade:
Looks good and my build isn't broken :)

PR #51, opened 2015-02-17
Unified Input

randomPoison:
- Introduced new context-based control scheme, where LMB does cutting and giving resources, and LMB does gliding and climbing.
- All input is now done through the `PlayerControls` script.
- Started splitting up the utils scripts.


Assets/Scenes/Garden.unity
```diff
@@ -57021,6 +57005,7 @@ MonoBehaviour:
   _leanTowardsSurface: .5
   _glideHorizontalSpeed: 5
   _glideDescentRate: 1
+  _glideTurnRate: .5
 --- !u!114 &1215378093
 MonoBehaviour:
   m_ObjectHideFlags: 0
```
brickman1444, line 57011: What are all these scene changes for? They will be squashed.

Assets/Scripts/Actors/ActorComponents/ActorCamera.cs
```diff
@@ -90,10 +90,10 @@ void CameraControl()
 
 			cam.transform.RotateAround(transform.position, 
 			                           cam.transform.right, 
-			                           Input.GetAxis("Mouse Y" + WadeUtils.platformName) * rotSpeed.y);
+			                           Input.GetAxis("Mouse Y" + PlatformUtils.platformName) * rotSpeed.y);
 			cam.transform.RotateAround(transform.position, 
 			                           cam.transform.up, 
-			                           Input.GetAxis("Mouse X" + WadeUtils.platformName) * rotSpeed.x);
+			                           Input.GetAxis("Mouse X" + PlatformUtils.platformName) * rotSpeed.x);
```
brickman1444: Make InputUtils.GetPlatformAxis(string name). Shouldn't be too hard and will wrap up this ugly repeated code.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -24,8 +24,11 @@ public List<BuddyTag> buddies
 	int resourceIndex = 0;
 	GameObject heldResource;
 
-	[SerializeField] LayerMask buddyLayer = 0;
-	[SerializeField] float maxGiveDistance = 2f;
+	[SerializeField] LayerMask _buddyLayer = 0;
+	public LayerMask buddyLayer
```
brickman1444, line 32: Does this make it effectively readonly?

Assets/Scripts/Cutting/Cutting.cs
```diff
@@ -4,58 +4,63 @@
...
 	{
 		cuttableComponent.Cut( _actorStats.GetStatValue( Stat.Cutting ) );
 	}
+
+	void CreateVisualEffect()
+	{
+		Vector3 camForward = Camera.main.transform.forward;
+		camForward.Set( camForward.x, 0, camForward.z );
+
+		actor.actorPhysics.OverrideLook( camForward, 0.5f );
```
brickman1444: What's this 0.5f doing? and if it's a speed why isn't it in the inspector?

Assets/Scripts/Cutting/Cutting.cs
```diff
@@ -4,58 +4,63 @@
...
 
 	void Cut( Cuttable cuttableComponent )
 	{
 		cuttableComponent.Cut( _actorStats.GetStatValue( Stat.Cutting ) );
 	}
+
+	void CreateVisualEffect()
+	{
+		Vector3 camForward = Camera.main.transform.forward;
+		camForward.Set( camForward.x, 0, camForward.z );
```
brickman1444: Why not just set the y value to 0?

Assets/Scripts/PlayerControls.cs
```diff
@@ -4,16 +4,24 @@
...
 public class PlayerControls : MonoBehaviour
 {
+	[Tooltip( "The distance forward from the camera's position to check for objects that can be interacted with." )]
+	[SerializeField] float _interactCheckDistance = 5.0f;
+	[SerializeField] float _interactCheckRadius = 0.2f;
+
 	PlayerActor _actor;
 	ActorPhysics _actorPhysics;
+	Cutting _cutting;
+	bool _holdButtonDown = false;
```
brickman1444, line 14: Is "Hold" for climbing and gliding?

Assets/Scripts/PlayerControls.cs
```diff
@@ -148,11 +178,55 @@ void GroundMovement()
...
+	 * items to interact with.
+	 * 
+	 * This performs RaycastAll, and returns the closest collision.
+	 */
+	bool RaycastForward( out RaycastHit closestHit )
+	{
+		Vector3 camPos = Camera.main.transform.position;
+		Vector3 camForward = Camera.main.transform.forward;
+
+		Debug.DrawRay( camPos, camForward * _interactCheckDistance, Color.yellow, 1.0f, false );
```
brickman1444, line 192: Do you want to keep this in?

Assets/Scripts/PlayerControls.cs
```diff
@@ -148,11 +178,55 @@ void GroundMovement()
...
+
+		RaycastHit[] hits = Physics.SphereCastAll(
+			new Ray( camPos, camForward ),
+			_interactCheckRadius,
+			_interactCheckDistance,
+			_cutting.cuttableLayer | _actor.actorResources.buddyLayer );
+
```
brickman1444, line 199: tabs/spaces

Assets/Scripts/PlayerControls.cs
```diff
@@ -148,11 +178,55 @@ void GroundMovement()
...
+		if ( hits.Length == 0 )
+		{
+			closestHit = new RaycastHit();
```
brickman1444, line 202: Why are you doing this instead of setting it to null?

Assets/Scripts/PlayerControls.cs
```diff
@@ -148,11 +178,55 @@ void GroundMovement()
...
+			return false;
+		}
+
+		closestHit = hits[0];
+		float closestDistance = ( camPos - closestHit.point ).sqrMagnitude;
```
brickman1444, line 207: campos? why not the character's position?

Assets/Scripts/PlayerControls.cs
```diff
@@ -148,11 +178,55 @@ void GroundMovement()
...
+	 * Retrives the raw input values, with horizontal mapped to X
+	 * and vertical mapped to Z.
+	 */
+	Vector3 GetMoveInput()
 	{
-		return new Vector3( Input.GetAxis( "Horizontal" + WadeUtils.platformName ),
+		return new Vector3( Input.GetAxis( "Horizontal" + PlatformUtils.platformName ),
 		                    0.0f,
-		                    Input.GetAxis( "Vertical" + WadeUtils.platformName ) );
+		                    Input.GetAxis( "Vertical" + PlatformUtils.platformName ) );
```
brickman1444: Remember to update these too if you make that input wrapper function

Assets/Scripts/Utilities/InputUtils.cs
```diff
@@ -0,0 +1,37 @@
+﻿using UnityEngine;
+using System.Collections;
+
+public static class InputUtils
+{
+	// TODO ask Chris what this is even for
+	public static void CheckForController()
```
brickman1444, line 7: @cjacobwade what even is this?

Assets/Scripts/Utilities/WadeUtils.cs
```diff
@@ -7,55 +7,11 @@
...
-		else
-		{
-			platformName = "_MOUSE";
-		}
-	}
-
 	public static float SMALLNUMBER = 0.00000001f;
 	public static float LARGENUMBER = 100000000f;
 
 	public static float DUALINPUTMOD = 0.7071f;
```
brickman1444, line 13: probably move this to InputUtils and rename it with underscores

Assets/Scripts/Utilities/WadeUtils.cs
```diff
@@ -7,55 +7,11 @@
...
-			#endif
-		}
-		else
-		{
-			platformName = "_MOUSE";
-		}
-	}
-
 	public static float SMALLNUMBER = 0.00000001f;
 	public static float LARGENUMBER = 100000000f;
```
brickman1444, line 11: move to mathutils and rename with underscores. And maybe rewrite in scientific notation.

Assets/Scripts/Utilities/InputUtils.cs
cjacobwade, line 7: If you put this in an update loop somewhere it will change the platformName to match whether a controller is plugged in or not.

Assets/Scripts/PlayerControls.cs
cjacobwade, line 202: RaycastHit can't have null value.

Assets/Scripts/PlayerControls.cs
cjacobwade, line 207: It feels better to check this from the camera direction rather than where the player is facing. We should probably turn the player towards a buddy when giving them stuff.

Assets/Scenes/Garden.unity
randomPoison, line 57011: They're automatic scene changes that are the result of changing what members the components have. It doesn't matter if they're squashed because the values are preserved in the prefabs.

Assets/Scripts/Actors/ActorComponents/ActorCamera.cs
randomPoison: Done. I named it `InputUtils.GetAxis()` to keep it consistent with the built in input methods.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
randomPoison, line 32: Yeah, it's readonly to scripts, but can still be set from the inspector. This is a pattern that we should be using fairly regularly.

Assets/Scripts/Cutting/Cutting.cs
randomPoison: It's the amount of time that the actor's look direction will be overridden. the hard-coded values was for testing purposes, I've moved it to be in the inspector.

Assets/Scripts/Cutting/Cutting.cs
randomPoison: Good question, I'll change it. BTW, you're the one who wrote that originally.

Assets/Scripts/PlayerControls.cs
randomPoison, line 14: Yeah. My logic is that the climbing and gliding buttons are held down during use, while the cutting and giving actions only require you to press the button once, so I've called the two actions "Hold" and "Use".

Assets/Scripts/PlayerControls.cs
randomPoison, line 192: Yeah, I think it's probably useful. If it get's too annoying we can take it out.

Assets/Scripts/PlayerControls.cs
randomPoison, line 199: This is correct. I want the arguments to be on a new line (and indented one more level), because if I kept them in line with method call the line would get too long.

Assets/Scripts/PlayerControls.cs
randomPoison, line 207: Good point, I'll add that.

Assets/Scripts/PlayerControls.cs
randomPoison: Done

Assets/Scripts/Utilities/WadeUtils.cs
randomPoison, line 13: I'm going to hold off on this, that can be done as part of a bigger push to clean up the utils.

Assets/Scripts/Utilities/InputUtils.cs
randomPoison, line 7: This isn't necessary because you can setup the input in such a way that it handles mouse vs joystick correctly.

PR #52, opened 2015-02-17
Buddy gameplay

brickman1444:
Some of this is bad but I needed to get it in.

After this is merged in the following needs to be done to the scene to avoid catastrophe:
1. Delete the Stamina Bar object under UI Canvas
2. Replace it with the Skills Bar prefab from the prefabs/UI folder
3. Add in bushes to show off the cutting
4. Select a valid stat type from the drop down on the Buddy Stats component of the Buddy that starts in the scene
5. Change the material of the buddy in the scene's body to match the skill selected. These are in the Materials folder and called "Buddy Gliding", "Buddy Cutting" and "Buddy Stamina"
6. Place 2 resource holders filled with buddy eggs in good places in the scene. Make sure each resource holder contains a different flavored egg and don't repeat the flavor of the buddy that starts in the scene
7. Drag the player into the Starting God field on the Buddy Stats component of the Buddy that starts in the scene

This should be in for the showbuild on 2/17. Text me/message me on slack if I don't respond to a comment on github while I'm at work.


Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -116,7 +120,21 @@ void SpawnBuddy()
 		BuddyStats newBuddy = ( Instantiate( buddyItemData.buddyPrefab,
 		                                     transform.position + transform.forward * 3.0f,
 		                                     Quaternion.identity ) as GameObject ).GetComponent<BuddyStats>();
```
cjacobwade, line 122: Weird formatting here and also magic number.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -116,7 +120,21 @@ void SpawnBuddy()
...
+		MeshRenderer[] childRenderers = newBuddy.gameObject.GetComponentsInChildren<MeshRenderer>();
+
+		foreach ( MeshRenderer meshRenderer in childRenderers )
+		{
+			if ( meshRenderer.gameObject.name == "Body" )
+			{
+				// buddyItemData.prefab.GetComponentInChildren<MeshRenderer>() was not working
+				MeshRenderer render = buddyItemData.prefab.transform.FindChild( "Sphere" ).gameObject.GetComponent<MeshRenderer>();
+				meshRenderer.material = render.material;
+			}
```
cjacobwade, line 135: Is this changing the buddies color to match the resource color? We should store a reference or make a getter in buddy item data to it's material.

Assets/Scripts/Actors/BuddyStats.cs
```diff
@@ -3,18 +3,32 @@
...
+		get { return _owner; }
+		set
+		{
+			_owner = value;
+			if ( _owner )
+			{
+				_actorStats = _owner.gameObject.GetComponent<PlayerActorStats>();
+			}
+		}
+	}
```
cjacobwade, line 31: Could we just replace this with PlayerActorStats since the tag aspect is only being used to get PlayerActorStats?

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
cjacobwade, line 135: Also, using render.material changes the prefab color rather than the instanced version used during the game. Instead use render.sharedMaterial

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
randomPoison, line 122: The formatting here is technically fine, though I'd move the `Instantiate()` call to a new line to keep it from getting overly long.

PR #55, opened 2015-02-21
DisplayOnly Attribute

brickman1444:
Attribute for only displaying a variable inspector without allowing the user to edit it. Good for preventing confusion of what's supposed to be edited.

I wish there were a more generic way to display the data but it looks like you have to just figure out what type it is. If you call toString() on the property then it just says "UnityEngine.SerializedProperty"


brickman1444:
Updated with combined version.

randomPoison:
http://answers.unity3d.com/questions/489942/how-to-make-a-readonly-property-in-inspector.html

That's where I got my original version of the attribute. It has some code for the height, see if that fixes the problem with things overlapping. If not then you're good to merge in anyway.

PR #56, opened 2015-02-22
Buddy Stats Work + Showbuilds + Physics Cleanup

randomPoison:
This includes Zac's changes to the buddy stats and my work on the physics system. Good luck reviewing, because the changes are myriad.
## Physics Changes
- Removed rolling state.
- Refactored the physics state machine to use proper state objects rather than delegates. This gives finer grain control over transitioning between states, allowing us to ensure that things always happen when entering or leaving a state.
- State transitions should no longer take place within `ActorPhysics` itself, the client states should manage that themselves.
- To facilitate this, all the `XCheck()` methods now return a value that represents the result of the check.
- `ActorPhysics` may still check internally what state it's in to do state-specific actions.
- `ActorPhysics` no longer deals with animation states, the client code must now do that.
- There are now `StartGliding()`, `StopGliding()`, `StartClimbing()`, and `StopClimbing()` methods for helping `ActorPhysics` with state transitions. These must be called by client code when transitioning to/from those states.


Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -19,12 +18,11 @@ public sealed class ActorPhysics : ActorComponent
 
 	// States
 	#region States
-	public delegate void ActorStateMethod();
 
 	[ReadOnly, SerializeField]
 	ActorStates _currentState = ActorStates.Jumping;
-	ActorStateMethod CurrentStateMethod;
-	Dictionary<ActorStates, ActorStateMethod> _stateMethodMap = new Dictionary<ActorStates, ActorStateMethod>();
+	PhysicsState _currentStateMethod = new DefaultState();
```
randomPoison: This shouldn't be called `_currentStateMethod` anymore, just `_currentState`.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -637,3 +561,17 @@ void EndJumpCheckDelay()
...
 }
+
+public abstract class PhysicsState
+{
+	public abstract void Enter();
+	public abstract void Update();
+	public abstract void Exit();
+}
+
+public class DefaultState : PhysicsState
```
randomPoison: `DefaultState` should go inside `ActorPhysics` the same way that the player's states go inside `ActorControls`.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -71,7 +71,30 @@ void CheckScroll()
...
+				canUse = true;
+			}
+		}
+
+		return canUse;
+	}
+
+	public void UseItem()
+	{
+		if ( !( heldResources[resourceIndex] is ResourceData ) )
```
randomPoison: ~~Change to `is BuddyItemData`.~~

Actually, change this to an assert. This shouldn't be called if value isn't correct, so we don't want to just silently walk over the mistake.

Assets/Editor/BuildUtils.cs
```diff
@@ -0,0 +1,32 @@
...
+using System;
+using UnityEditor;
+
+public class BuildUnityPlayer : MonoBehaviour
+{
+	[MenuItem( "Export/Build All" )]
+	public static void PerformBuild()
+	{
+		// the scenes we want to include in the build
+		string[] scenes = {"Scenes/BlockOut.unity"};
```
brickman1444: what is this?

Assets/Editor/BuildUtils.cs
```diff
@@ -0,0 +1,32 @@
...
+
+		// build for windows stand alone
+		string macStandAloneBuildName = buildName + "_OSX.exe";
+		EditorUserBuildSettings.SwitchActiveBuildTarget( BuildTarget.StandaloneOSXUniversal );
+		BuildPipeline.BuildPlayer( scenes, macStandAloneBuildName, BuildTarget.StandaloneOSXUniversal, BuildOptions.None );
+
+		// build for windows stand alone
+		string linuxStandAloneBuildName = buildName + "_LINUX.exe";
+		EditorUserBuildSettings.SwitchActiveBuildTarget( BuildTarget.StandaloneLinuxUniversal );
+		BuildPipeline.BuildPlayer( scenes, linuxStandAloneBuildName, BuildTarget.StandaloneLinuxUniversal, BuildOptions.None );
```
brickman1444, line 30: you could wrap up these three repetitive sections in a function that takes in a build name and a buildtarget.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -122,7 +142,21 @@ void SpawnBuddy()
...
+		newBuddy.statType = buddyItemData.stat;
+
+		MeshRenderer[] childRenderers = newBuddy.gameObject.GetComponentsInChildren<MeshRenderer>();
+
+		foreach ( MeshRenderer meshRenderer in childRenderers )
+		{
+			if ( meshRenderer.gameObject.name == "Body" )
+			{
+				// buddyItemData.prefab.GetComponentInChildren<MeshRenderer>() was not working
+				MeshRenderer render = buddyItemData.prefab.transform.FindChild( "Sphere" ).gameObject.GetComponent<MeshRenderer>();
```
randomPoison: Eeeeewwwwww. This is nasty.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -19,12 +18,10 @@ public sealed class ActorPhysics : ActorComponent
 
 	// States
 	#region States
-	public delegate void ActorStateMethod();
-
 	[ReadOnly, SerializeField]
-	ActorStates _currentState = ActorStates.Jumping;
-	ActorStateMethod CurrentStateMethod;
-	Dictionary<ActorStates, ActorStateMethod> _stateMethodMap = new Dictionary<ActorStates, ActorStateMethod>();
+	ActorStates _currentStateType = ActorStates.Jumping;
```
brickman1444: Rename to physicsstatetype

Assets/Scripts/Actors/ActorComponents/PlayerActorStats.cs
```diff
@@ -24,6 +24,10 @@ public class StatObject
 	public Image currentImage = null;
 	public Image maxImage = null;
 	public float statToScaleRatio = 0.0f;
+	public float rechargeDelayTime = 0.0f; // Seconds
+
+	float _rechargeTimer = 0.0f; // Made a property so that it doesn't show up in the inspector where it will be confusing but will show in debug mode.
+	public float rechargeTimer { get { return _rechargeTimer; } set { _rechargeTimer = value; } }
```
randomPoison: This would probably be an appropriate use for the fnacy new `[ReadOnly]` attribute.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -102,13 +99,6 @@ public float rollMoveSpeed
 	bool _isOnGround = false;
```
brickman1444, line 106: consider making readonly so it shows up for debugging

Assets/Scripts/Actors/BuddyStats.cs
```diff
@@ -3,18 +3,29 @@
...
-	[SerializeField] float decreaseResourcesTime = 60.0f; //seconds
-	[SerializeField] float currResourceTimer = 0.0f;
+	public Stat statType = Stat.Invalid;
 	[SerializeField] int startingApples = 10;
 	[SerializeField] int apples = 0;
-
-	public GodTag owner = null;
+	[SerializeField] GodTag _startingOwner = null;
 
 	PlayerActorStats _actorStats = null;
```
randomPoison: This is poorly named, it should be something like `_ownerStats`. As it is it implies that it refers to the buddy's `PlayerActorStats` component.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -276,7 +255,10 @@ public bool ClimbCheck()
 					}
 				}
 
-				StartClimbing( nearestCol.GetComponent<ClimbableTag>() );
+				ClimbableTag climbTag = nearestCol.GetComponent<ClimbableTag>();
+				_climbTag = climbTag;
+				_climbSurface = _climbTag.GetComponent<Transform>();
```
brickman1444, line 265: what's the point of climbTag instead of _climbTag?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -497,36 +426,30 @@ void InitializeStateMethodMap()
...
-		bumper.gameObject.SetActive( false );
-		climbBumper.gameObject.SetActive( true );
-
-		if ( !actor.animator )
-		{
-			actor.animator.SetBool( "isClimbing", true );
-		}
+		_currentStateType = toState;
+		_currentState.Exit();
+		_currentState = _stateMap[_currentStateType];
```
brickman1444, line 458: personally I'd replace the _currentStateType here with toState just in case the order of these statements ever gets switched around.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -156,22 +146,22 @@ void FixedUpdate()
...
 		{
-			_stateMethodMap[state] = method;
+			_stateMap[state] = method;
 		}
 		else
 		{
-			_stateMethodMap.Add( state, method );
+			_stateMap.Add( state, method );
 		}
 	}
```
cjacobwade, line 173: It's not really clear what PhysicsState is here. Maybe rename the type to explain that it's a method or change method to something else if it isn't.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -179,43 +169,34 @@ public bool GroundedCheck()
...
-		                    out hit,
-		                    _jumpCheckDistance,
-		                    _jumpLayer );
-		_isOnGround = ( hit.transform &&
-		                Vector3.Dot( hit.normal, Vector3.up ) > _minJumpDot );
-
-		if ( _isOnGround &&
-		     !_jumpCheck &&
-		     !IsInState( ActorStates.Climbing ) )
+		if ( !_jumpCheck )
```
cjacobwade: This logic seems unclear. Why would we check for a jump if jumpCheck is false?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -276,7 +255,10 @@ public bool ClimbCheck()
 					}
 				}
 
-				StartClimbing( nearestCol.GetComponent<ClimbableTag>() );
+				ClimbableTag climbTag = nearestCol.GetComponent<ClimbableTag>();
+				_climbTag = climbTag;
```
cjacobwade: Why isn't this a direct assignment?

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -88,29 +111,26 @@ public void CheckUseItem( RaycastHit hitInfo )
...
+			_playerActor.physics.OverrideLook(
+				buddyStats.GetComponent<Transform>().position - GetComponent<Transform>().position,
+				_lookOverrideDuration );
 		}
 	}
 
 	void GiveResource( BuddyStats buddyStats )
 	{
-		buddyStats.GiveResource( (actor as PlayerActor).actorStats, (ResourceData)heldResources[resourceIndex] );
+		buddyStats.GiveResource( (actor as PlayerActor).stats, (ResourceData)heldResources[resourceIndex] );
```
brickman1444, line 128: consider changing to ( ( PlayerActor ) actor ).stats

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -122,7 +142,21 @@ void SpawnBuddy()
 		BuddyStats newBuddy = ( Instantiate( buddyItemData.buddyPrefab,
 		                                     transform.position + transform.forward * 3.0f,
```
brickman1444: This 3.0f should probably be in the inspector.

Assets/Scripts/Actors/ActorComponents/PlayerActorStats.cs
brickman1444: take out this property and tag _rechargeTimer with [ReadOnly]

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -319,54 +320,7 @@ public void ClimbSurface( Vector3 movement )
...
-		_rollCooldownTimer += Time.deltaTime;
-	}
-
-	public void RollMovement()
-	{
-		MoveAtSpeed( _moveVec.normalized, _rollMoveSpeed );
-	}
-
-	public void JumpCheck()
+	public bool JumpCheck()
```
cjacobwade, line 328: There might be a more descriptive name we can do here. This name somewhat ambiguous between Can I Jump and Perform Jump. I suggest TryJump

brickman1444:
I read through everything.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade: We should rename this to DefaultPhysicsState. Since it inherits from Physics state it's not like we're going to reuse it for anything else.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -71,7 +71,30 @@ void CheckScroll()
...
+
+	public void UseItem()
+	{
+		if ( !( heldResources[resourceIndex] is ResourceData ) )
+		{
+			SpawnBuddy();
+		}
+	}
+
+	public void UseItemWithTarget( RaycastHit hitInfo )
```
cjacobwade, line 89: These behaviours are all specific to Resources so we should probably rename all of them them UseResourceEtc

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
```diff
@@ -88,29 +111,26 @@ public void CheckUseItem( RaycastHit hitInfo )
 
 	void CheckGiveResources( RaycastHit hitInfo )
 	{
-		if ( hitInfo.transform )
-		{
-			BuddyStats buddyStats = hitInfo.transform.GetComponent<BuddyStats>();
+		BuddyStats buddyStats = hitInfo.transform.GetComponent<BuddyStats>();
```
cjacobwade, line 108: Do we know that hitInfo will always have content when it gets here? Going to run into errors otherwise

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
cjacobwade: no

Assets/Scripts/Actors/BuddyStats.cs
```diff
@@ -30,19 +41,9 @@ void Awake()
...
-	}
-
-	public void SetGod( GodTag _godTag )
-	{
-		if(_godTag)
-		{
-			owner = _godTag;
-			_actorStats = _godTag.gameObject.GetComponent<PlayerActorStats>();
-		}
+		owner = _startingOwner; // For owners set in the inspector
```
cjacobwade: Can we just get the god in the scene for this so we don't have to manually assign it for every buddy in the scene? We can change it later if we get to other gods.

Assets/Scripts/Actors/PlayerActor.cs
```diff
@@ -7,22 +7,34 @@ public class PlayerActor : Actor
...
 	{
 		get { return _actorStats; }
 	}
 
+	public PlayerControls controls
+	{
+		get { return _controls; }
+	}
+
+	public Cutting cutting
```
cjacobwade: Consider renaming this to keep with the naming scheme of PlayerX

Assets/Scripts/Actors/PlayerActor.cs
```diff
@@ -7,22 +7,34 @@ public class PlayerActor : Actor
...
 	}
 
-	public PlayerActorInventory actorResources
+	public PlayerActorInventory inventory
 	{
 		get { return _playerActorResources; }
 	}
 
-	public PlayerActorStats actorStats
+	public PlayerActorStats stats
```
cjacobwade: We can probably rename these to PlayerInventory and PlayerStats since they're specific to the player.

Assets/Scripts/PlayerControls.cs
```diff
@@ -9,18 +9,28 @@ public class PlayerControls : MonoBehaviour
...
 
 	PlayerActor _actor;
-	ActorPhysics _actorPhysics;
-	Cutting _cutting;
 	
 	Button _holdButton = new Button( "Hold" );
+	public Button holdButton
+	{
+		get { return _holdButton; }
+	}
```
cjacobwade: I still feel like this is confusing especially in the context of buttons. Consider grasp, clutch or grip.

Assets/Scripts/PlayerControls.cs
```diff
@@ -38,123 +48,212 @@ void Update()
...
+			{
+				player.physics.ChangeState( ActorStates.Falling );
+			}
 		}
+
+		public override void Exit() { }
 	}
 
-	void Gliding()
+	public class Gliding : PhysicsState
```
cjacobwade: Can we put all of these Physics states in a namespace or something? Seems like these class names could potentially get used for something else way down the line.

Assets/Scripts/PlayerControls.cs
```diff
@@ -38,123 +48,212 @@ void Update()
...
+			player.animator.SetBool( "isGliding", true );
 		}
 
-		if ( !_actorPhysics.GroundedCheck() &&
-		     _holdButton &&
-		     _actor.actorStats.CanUseStat( Stat.Gliding ) )
+		public override void Update()
 		{
-			_actorPhysics.GlideMovement( GetMoveDirection() );
+			if ( player.controls.jumpButton.down &&
```
cjacobwade: Didn't you add that override thing to this? so we can just have player.controls.jumpButton? Or if this is like just pressed then it's unclear

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
```diff
@@ -179,43 +169,34 @@ public bool GroundedCheck()
...
+			// spherecast down to detect when we're on the ground
+			RaycastHit hit;
+			Physics.SphereCast( transform.position + Vector3.up,
+			                    _jumpCheckRadius,
+			                    -Vector3.up * _jumpCheckDistance,
+			                    out hit,
+			                    _jumpCheckDistance,
+			                    _jumpLayer );
+			_isOnGround = ( hit.transform &&
+			                Vector3.Dot( hit.normal, Vector3.up ) > _minJumpDot );
```
cjacobwade, line 190: Using the hit.normal directly could lead to not being able to jump on some weird mesh surface if there's a teeny face that's steep. That's probably a good reason to start making collision meshes...

Assets/Scripts/PlayerControls.cs
```diff
@@ -38,123 +48,212 @@ void Update()
...
 		}
 	}
+
+	void SetupStateMethodMap()
+	{
+		_actor.physics.RegisterStateMethod( ActorStates.Jumping,  new Jumping( _actor ) );
+		_actor.physics.RegisterStateMethod( ActorStates.Falling,  new Jumping( _actor ) );
+		_actor.physics.RegisterStateMethod( ActorStates.Grounded, new Grounded( _actor ) );
+		_actor.physics.RegisterStateMethod( ActorStates.Climbing, new Climbing( _actor ) );
+		_actor.physics.RegisterStateMethod( ActorStates.Gliding,  new Gliding( _actor ) );
```
cjacobwade: This isn't registering a state method anymore, it's just doing states

Assets/Scripts/PlayerControls.cs
```diff
@@ -188,7 +287,7 @@ bool RaycastForward( out RaycastHit closestHit )
 			new Ray( camPos, camForward ),
```
cjacobwade: Consider a more descriptive name for this method. CheckForHitInteractable maybe?

cjacobwade: Consider a more descriptive name for this method. CheckForHitInteractable maybe?

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 106: I think this is something that's better suited for just enabling the debug inspector.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 265: Yeah, that was dumb. I've fixed it.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 173: Are you saying add a comment to the `PhysicsState` class explaining its purpose? It's not clear from the line you commented on what you're referring to.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: `_jumpCheck` is a poorly named variable. It's actually the late jump delay. I'll rename it to make things more clear.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison: Fixed.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade, line 173: I'm saying that using the variable name method when adding to _stateMap doesn't make sense because you changed it to a PhysicsState object

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
randomPoison, line 128: Actually, if `actor` needs to be a `PlayerActor`, it should just be a `PlayerActor`. I think this problem stems from the `ActorComponent` base class, which should probably be removed.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
randomPoison: done

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 328: I see what you mean, but I think it's consistent within the `ActorPhysics` system. All of the `XCheck()` methods determine if action X can happen, perform the necessary actions as needed, and then return the result of the attempt. I think the bigger problem is that we're both checking if the action is possible and performing the action in the same method. With the latest round of changes the more appropriate way of doing this would be to have a separate `StartJumping()` method, something in the ilk of `startClimbing()` and `StartGliding()`.

Assets/Scripts/Actors/PlayerActor.cs
randomPoison: I actually changed these to rename remove the "player" prefix from their name. In the context they're being used you know that you're accessing the members of `PlayerActor`, so doing `PlayerActor.playerPhysics` is redundant compared to `PlayerActor.physics`. I figured this is inline with how Unity handles things like this, such as how `Random.RandomRange()` is deprecated in favor of `Random.Range()`.

Assets/Scripts/Actors/PlayerActor.cs
randomPoison: Yeah, good point.

Assets/Scripts/PlayerControls.cs
randomPoison: Yeah, good point. Fixed.

Assets/Scripts/PlayerControls.cs
randomPoison: This is like `Input.GetButtonDown()`, so it's only true if the button was just pressed this frame. I'm not sure how we can make this more clear while sticking the the Unity standards.

Assets/Scripts/PlayerControls.cs
randomPoison: These are all within the `PlayerControls` class, so there's no worry about name conflicts.

Assets/Scripts/PlayerControls.cs
randomPoison: The button is also for gliding, so  I don't feel like those work any better. It makes sense if you think of the two inputs as "use" and "hold", though I get why that looks weird at first glance. I'm up to changing it if you have something that describes it better.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
cjacobwade, line 328: I agree that it doesn't fit, but that's because they're all ambiguous in that way. Using those StartX methods is good, but XCheck should still be more readable, I think.

Assets/Scripts/Actors/BuddyStats.cs
randomPoison: Yeah, I agree that's a better way of doing it for now.

Assets/Scripts/Actors/ActorComponents/PlayerActorInventory.cs
randomPoison, line 108: It should always have data by this point, but we can add an assert to check if that's the case.

Assets/Scripts/Actors/ActorComponents/ActorPhysics.cs
randomPoison, line 328: Okay. I'll split them up into things like `CheckJump()` and `DoJump()` to keep things consistent and decoupled.

randomPoison, line 328: Also I'm going to do that as part of the next pull request, since I'll have to do further physics refactoring anyway.

PR #59, opened 2015-02-23
Day/Night Mechanics

brickman1444:
Buddies all lose resources at midnight. The player is teleported back to the start at midnight. This code should actually be good.

Scene Changes required:
BuddyManager, DayCycleManager and RenderSettingsManager are all prefabs. Make sure that they are present in the scene and that the instances in the scenes are instances of these prefabs.
In Prefabs/UI is a prefab called Black Overlay. Put this in the scene under the UICanvas. Drag a reference to this object into the DayNightCycleManager for the Midnight Overlay slot.
Drag a reference to the play into the DayNightCycleManager for the PlayerControls slot.
Double check that the RenderSettingsManager has a reference to the directional light in the scene.


randomPoison:
Make `BuddyManager`, `DayCycleManager`, and `RenderSettingsManager` should all be on the same prefab. All managers like this should go on a single "Level Manager" prefab.

cjacobwade:
Something to consider, if those managers need to create an object for some reason it makes sense to make that object a child for scene organization's sake. If all managers are the same object, it's children may end up a muddle of mixed things.

brickman1444:
Good point, I think you could still combine all the non object spawning managers into one object, though. and possibly making the object spawning ones child objects.

brickman1444:
Let me know once you guys review the rest of this.

cjacobwade:
Yeah, that's a good way to go about that. Checking the rest now.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -4,11 +4,14 @@
 public class BuddyStats : MonoBehaviour
 {
 	public Stat statType = Stat.Invalid;
-	[SerializeField] int startingApples = 10;
-	[SerializeField] int apples = 0;
+	[SerializeField] bool canDecreaseStat = true;
+	[SerializeField] int startingResources = 10;
```
cjacobwade: A more descriptive/concise name for this might be initResourceCount or startingResourceCount. This could be misconstrued as and array of resources prefabs as is

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -4,11 +4,14 @@
...
+	[ReadOnly("Current Resources")]
+	[SerializeField] int resources = 0;
```
cjacobwade: Add underscore to these. Also this name should be more descriptive, I recommend currentResourceCount.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -4,11 +4,14 @@
 public class BuddyStats : MonoBehaviour
 {
 	public Stat statType = Stat.Invalid;
-	[SerializeField] int startingApples = 10;
-	[SerializeField] int apples = 0;
+	[SerializeField] bool canDecreaseStat = true;
```
cjacobwade: Maybe call this canDecreasePlayerStat?

Assets/Scripts/BuddyManager.cs
```diff
@@ -0,0 +1,23 @@
...
+	}
+
+	public static void DecrementAllBuddyResources()
+	{
+		foreach ( BuddyStats buddyStat in instance._buddyStats )
+		{
+			buddyStat.DecrementResources();
+		}
+
+		instance._buddyStats.RemoveAll( buddyStat => buddyStat.isAlive == false ); // Predicate statement
```
cjacobwade, line 26: Dang this is cool. I didn't really understand the use of lambdas/predicates in c# until seeing this an looking it up. Thought it only worked w/ LINQ

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -9,6 +9,15 @@ public class PlayerControls : MonoBehaviour
 	[SerializeField] float _interactCheckRadius = 0.2f;
 
 	PlayerActor _actor;
+
+	[Tooltip( "Probably straight up" )]
+	[SerializeField] Vector3 _respawnOffset = new Vector3();
+	[ReadOnly( "Respawn Position" )]
+	[SerializeField] Vector3 _respawnPosition = new Vector3();
+	Quaternion _respawnRotation = new Quaternion();
```
cjacobwade: I don't know if this is costly to have new Quaternion/Vector3 but  Quaternion.Identity or Vector3.Zero might be more readable.

cjacobwade:
In playerControls in RaycastForward there's a few instances of Camera.main.transform which is a wrapper for GetObjectWithTag("MainCamera").GetComponent<Transform>() which is expensive as heck. I think it's a good idea to cache the camera here and in other scripts that use the camera somewhere.

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -0,0 +1,80 @@
...
+
+		if ( !_hasStartedMidnightOverlay && _dayCycleTimer > _blackOutStartTime )
+		{
+			StartMidnightOverlay();
+		}
+
+		if ( _dayCycleTimer > _dayCycleLength )
+		{
+			BuddyManager.DecrementAllBuddyResources();
+			_playerControls.Respawn();
```
cjacobwade: This seems a weird place to have reference to the player and buddy manager. We should consider having them subscribe to some EndOfDay callback.

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -0,0 +1,80 @@
...
+		Color color = _midnightOverlay.color;
+		color.a = 0.0f;
+		_midnightOverlay.color = color;
+	}
+
+	IEnumerator FadeInMidnightOverlay()
+	{
+		while ( true )
+		{
+			Color color = _midnightOverlay.color;
```
cjacobwade: Should be named overlayColor or something although context makes this clear.

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -0,0 +1,80 @@
...
+		_midnightOverlayCoroutine = StartCoroutine( FadeInMidnightOverlay() );
+	}
+
+	void EndMidnightOverlay()
+	{
+		_hasStartedMidnightOverlay = false;
+
+		StopCoroutine( _midnightOverlayCoroutine );
+
+		Color color = _midnightOverlay.color;
```
cjacobwade: Should be named overlayColor or something although context makes this clear.

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -21,97 +32,79 @@ public struct RenderSettingsData
...
+	public float daylightIntensity
+	{
+		get
+		{
+			return _daylightIntensity;
+		}
+		private set
+		{
+			_daylightIntensity = value;
+		}
```
cjacobwade: Get and set could be one line each, right? Not sure what our code guideline is for this. Readability would be fine, I think.

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -21,97 +32,79 @@ public struct RenderSettingsData
...
 	void Awake()
 	{
-		_curSkyboxMaterial = _currentRenderSettings.skyMaterial;
 		_curSkyboxTintPropertyID = Shader.PropertyToID( "_Tint" );
 	}
 
 	void Update()
 	{
-		_dayCycleTimer += Time.deltaTime;
+		daylightIntensity = Mathf.Cos( DayCycleManager.dayCycleTimer / DayCycleManager.dayCycleLength * 2 * Mathf.PI ) * -0.5f + 0.5f;
```
cjacobwade: 2 here should be 2f to save the implicit cast

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -21,97 +32,79 @@ public struct RenderSettingsData
...
 
-		timeOfDay = ( -Mathf.Cos( _dayCycleTimer/_dayCycleTime * 2f * Mathf.PI ) * 0.5f ) + 0.5f;
+		_currentTimeRenderSettings = TimeRenderSettings.Lerp( _currentRenderSettings.nightSettings, _currentRenderSettings.daySettings, daylightIntensity );
```
cjacobwade: The change to daylightIntensity rather than timeOfDay makes sense and is interesting. Do you think the concept of night/day time settings still make sense or should it be light/dark or something else?

brickman1444: I think Day/Night still makes the most sense. They are light and dark settings associated with a time of day. I think Day and NIght sums that up well.

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -21,97 +32,79 @@ public struct RenderSettingsData
...
 
-	public static void ChangeTargetRenderSettings( RenderSettingsData renderSettings, float shiftTime )
+	public static void TransitionRenderSettings( RenderSettingsData newRenderSettings, float shiftTime )
 	{
-		instance.IChangeTargetRenderSettings( renderSettings, shiftTime );
+		instance._TransitionRenderSettings( newRenderSettings, shiftTime );
 	}
 
-	void IChangeTargetRenderSettings( RenderSettingsData renderSettings, float shiftTime )
+	void _TransitionRenderSettings( RenderSettingsData newRenderSettings, float shiftTime )
```
cjacobwade: We can just move this into TransitionRenderSettings, right? Also not a fan of the _MethodName, but not sure what our code standard is for that.

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -21,97 +32,79 @@ public struct RenderSettingsData
...
-		_targetRenderSettings = renderSettings;
-
-		StopAllCoroutines();
-		StartCoroutine( GoToTargetRenderSettings( shiftTime ) );
+		StopAllCoroutines(); // TODO: Make this not a sledgehammer solution
+		StartCoroutine( TransitionRenderSettingsRoutine( newRenderSettings, shiftTime ) );
 	}
 
-	void ApplyRenderSettings()
+	void ApplyTimeRenderSettings()
```
cjacobwade: Rather than using this method, I think it would be better if we make _currentRenderSettings a property and make its Setter do this bit. Then we're only calling this when things are changed and we can move it out of update.

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: I'll take a look at how it is if it's all moved into the static method. I don't like the _MethodName either but I think it works very well for private instance functions on singletons.

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: That's a good idea. I don't think it will save a lot because most stuff is changing every frame but stuff like the sky material would only update when it changes.

cjacobwade:
Okay, all done

Assets/Scripts/Buddy/BuddyStats.cs
randomPoison: This is also poorly named since it's just a debug variable. It should be changed to something like `disableStatEffects`, and you should use a `[Header()]` attribute to split debug settings off from the main settings.

Assets/Scripts/Player/PlayerControls.cs
randomPoison: `Quaternion` and `Vector3` are structs, so they're functionally identical. That being said, it would still be a good idea to to use `Quaternion.Identity` and `Vector3.Zero` because they're more explicit about what values are being set.

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444: done

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444: done

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444: done

Assets/Scripts/Player/PlayerControls.cs
brickman1444: done

Assets/Scripts/Utilities/DayCycleManager.cs
brickman1444: done

brickman1444: done

Assets/Scripts/Utilities/DayCycleManager.cs
brickman1444: done

Assets/Scripts/Utilities/DayCycleManager.cs
brickman1444: done

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: done

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: done

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: done

Assets/Scripts/Utilities/RenderSettingsManager.cs
brickman1444: done

PR #60, opened 2015-02-23
Physics Cleanup

randomPoison:
The big thing here is that I removed all the check within `ActorPhysics` that were doing things like `if ( IsInState( whatever ) ) { do whatever }`. Now everything is done either in the client state machine, or in more appropriate methods based on when the action should be performed. I also removed `ActorPhysics.IsInState()` to ensure that this is avoided in the future.


Assets/Scripts/Actors/ActorPhysics.cs
```diff
@@ -460,6 +420,32 @@ public void ChangeState( PhysicsStateType toState )
...
+		Vector3 actualVelocity = transform.position - _lastPos;
+		actualVelocity.y = 0.0f;
+
+		Vector3 intendedVelocity = rigidbody.velocity;
+		intendedVelocity.y = 0.0f;
+
+		Vector3 lookVec =
+			( _overrideLook ?
+				_lookOverride :
+				Vector3.Lerp( actualVelocity, intendedVelocity, _lookIntentionWeight ) );
```
cjacobwade: This is probably worth breaking up for readability's sake

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -60,6 +60,60 @@ public Jumping( PlayerActor player )
...
+			player.physics.DoJump();
+		}
+
+		public override void Update()
+		{
+			if ( player.physics.GroundedCheck() )
+			{
+				if ( player.controls.jumpButton.down &&
+				     player.physics.JumpCheck() )
+				{
```
cjacobwade: This seems like a good place to actually execute our Jump if JumpCheck is true, right? Still seems a weird naming convention to me

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -120,6 +179,7 @@ public override void Update()
 			else
```
cjacobwade, line 179: When checking if we should climb before this, it's cheaper to do ClimbCheck last since you could early out if holdbutton isn't down or if we can't use stamina.

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -229,7 +285,15 @@ public override void Update()
 				}
 				else
 				{
-					player.physics.ChangeState( PhysicsStateType.Grounded );
+					if ( player.controls.jumpButton.down &&
```
cjacobwade: Do we want it to be possible to jump during this frame b/w gliding and landing? Seems weird but I guess it doesn't hurt

cjacobwade:
Good job on this. New system is super clean!

Assets/Scripts/Actors/ActorPhysics.cs
randomPoison: Fixed.

Assets/Scripts/Player/PlayerControls.cs
randomPoison, line 179: Fixed.

Assets/Scripts/Player/PlayerControls.cs
randomPoison: I've been trying to give the player as much opportunity to jump as possible, which includes allowing the player to jump again before they've quite hit the ground. I figure this is in the vein of the late jump delay. That said, I reworked the logic of this one to make it a bit clearer.

PR #62, opened 2015-02-24
Sprinting

brickman1444:
Sprinting physics state

Expands animator with blend tree for walk / run. If we get a run animation, there is now a slot for it. Currently it's just the walk animation sped up.

not sure why it's saying I wrote the whole player controls script...


cjacobwade:
Looks good.

PR #66, opened 2015-03-01
Day/Night Cycle Improvments

randomPoison:
- Added a fade in in the morning so that you don't see the player teleport.
- Added a fixed portion of the day/night cycle that stays night at the end.
- Made night settings for all renders darker.
- Made the skybox texture darker.
- Made the fade out effect last a fixed amount of time.


Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -58,7 +58,7 @@ void Awake()
 		DayCycleManager.RegisterEndOfDayCallback( Respawn );
 	}
 
-	void Update()
+	void FixedUpdate()
 	{
```
brickman1444, line 69: Why was this change made? Isn't input usually handled on Update() ?

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -4,55 +4,61 @@
...
+	public delegate void EndOfDayCallback();
+
 	public static float dayCycleLength { get { return instance._dayCycleLength; } }
 
+	[ReadOnly( "Current Time" ), Tooltip( "0 is midnight. 1/2 of Day Cycle Length is noon." ), SerializeField]
+	float _currentTime = 0f;
+
 	[Tooltip( "The length (in seconds) of a day" )]
-	[SerializeField] float _dayCycleLength = 60.0f;
+	[SerializeField] float _dayCycleLength = 300.0f;
```
brickman1444, line 15: You shouldn't be worried about updating this when you change the value in the inspector. Just keep it to a sane default value.

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -61,34 +67,64 @@ public static void RegisterEndOfDayCallback( EndOfDayCallback callback )
...
+	void StartMorningOverlay()
 	{
-		_hasStartedMidnightOverlay = false;
-
-		StopCoroutine( _midnightOverlayCoroutine );
-
-		Color overlayColor = _midnightOverlay.color;
-		overlayColor.a = 0.0f;
-		_midnightOverlay.color = overlayColor;
+		StartCoroutine( FadeOutMidnightOverlay() );
```
brickman1444: Are StartMidnightOverlay() and StartMorningOverlay() worth having? They both only have one line each. I think it'd be clearer to remove them

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -61,34 +67,64 @@ public static void RegisterEndOfDayCallback( EndOfDayCallback callback )
...
+	 * Used to fade to black at midnight.
+	 */
 	IEnumerator FadeInMidnightOverlay()
 	{
+		float elapsedTime = 0.0f;
+		while ( true )
+		{
+			if ( elapsedTime < _blackoutDuration )
+			{
+				elapsedTime += Time.deltaTime;
```
brickman1444, line 88: Do you want to be ticking inside this check?

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -61,34 +67,64 @@ public static void RegisterEndOfDayCallback( EndOfDayCallback callback )
...
+				Color overlayColor = _midnightOverlay.color;
+				overlayColor.a = elapsedTime / _blackoutDuration;
+				_midnightOverlay.color = overlayColor;
```
brickman1444, line 91: If the color can't be changed directly you should go ahead and write the extension methods so we can do:

_midnightOverlay.color = _midnightOverlay.color.SetA( elapsedTime / _blackoutDuration );

Assets/Scripts/Utilities/DayCycleManager.cs
```diff
@@ -61,34 +67,64 @@ public static void RegisterEndOfDayCallback( EndOfDayCallback callback )
...
 
+	/**
+	 * Used to fade to black at midnight.
+	 */
 	IEnumerator FadeInMidnightOverlay()
 	{
+		float elapsedTime = 0.0f;
+		while ( true )
+		{
+			if ( elapsedTime < _blackoutDuration )
```
brickman1444, line 86: Consider reformatting to 

```
elapsedTime += Time.deltaTime;
Color overlayColor = _midnightOverlay.color;
overlayColor.a = elapsedTime / _blackoutDuration;
_midnightOverlay.color = overlayColor;

if ( elapsedTime >= _blackoutDuration )
{
    yield return break;
}

yield return null;
```

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -3,23 +3,23 @@
...
-	public Color skyColor = Color.white;
-	public Color lightColor = Color.white;
-	public float lightIntensity = 0.5f;
-	public Color fogColor = Color.white;
-	public float fogDensity = 0.05f;
+	public Color skyColor;
+	public Color lightColor;
+	public float lightIntensity;
+	public Color fogColor;
+	public float fogDensity;
```
brickman1444, line 12: How does Unity fill these in when you create a new one in the inspector? 0's and whites?

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -96,7 +99,16 @@ void Awake()
 
 	void Update()
 	{
-		daylightIntensity = Mathf.Cos( DayCycleManager.dayCycleTimer / DayCycleManager.dayCycleLength * 2.0f * Mathf.PI ) * -0.5f + 0.5f;
+		float timeOfDay = DayCycleManager.currentTime / DayCycleManager.dayCycleLength;
```
brickman1444: I would put the night fraction adjustment in here.

```
timeOfDay = DayCycleManager.currentTime / ( DayCycleManager.dayCycleLength * ( 1.0f - _nightFraction) );
```

Then you compare it to 1.0f in the if statement and take out the division from the cosine line

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -77,12 +84,8 @@ TimeRenderSettings _currentTimeRenderSettingsProperty
 		}
 	}
 
-	[SerializeField] Light _dirLight = null;
-	int _curSkyboxTintPropertyID = 0;
-
 	[ReadOnly("Daylight Intensity"), Tooltip( "0 is midnight. 1 is noon." )]
-	[SerializeField]
-	float _daylightIntensity = 1.0f;
+	[SerializeField] float _daylightIntensity = 1.0f;
```
brickman1444, line 65: Why does this serialize? You never would really want to set it. It would be changed again the next frame.

Assets/Scripts/Utilities/RenderSettingsManager.cs
```diff
@@ -3,23 +3,23 @@
...
-	public float lightIntensity = 0.5f;
-	public Color fogColor = Color.white;
-	public float fogDensity = 0.05f;
+	public Color skyColor;
+	public Color lightColor;
+	public float lightIntensity;
+	public Color fogColor;
+	public float fogDensity;
 
 	public static TimeRenderSettings Lerp(TimeRenderSettings fromSettings, TimeRenderSettings toSettings, float t)
```
brickman1444: consider adding lerp methods to these structs that take in the output location by reference. It would save on copying memory around.

brickman1444:
done

Assets/Scripts/Player/PlayerControls.cs
randomPoison, line 69: This fixes the problem of one click giving multiple items. It might also help with some of the jumping problems, but I'm not sure. When we were reading input in `Update()` input was being read less often than physics was being updated in `FixedUpdate()` so 2 or 3 frames would be using the same input data. This fixes it for now, though we should seek out a better solution at some point.

Assets/Scripts/Utilities/DayCycleManager.cs
randomPoison, line 15: I was trying to change it to a sane default value.

Assets/Scripts/Utilities/DayCycleManager.cs
randomPoison: `StartMidnightOverlay()` is needed because it's called by an `Invoke()` (blech), but I can get rid of `StartMorningOverlay()`.

Assets/Scripts/Utilities/DayCycleManager.cs
randomPoison, line 88: It's functionally the same as ticking before the check, and I think this is a bit cleaner.

Assets/Scripts/Utilities/RenderSettingsManager.cs
randomPoison, line 12: Just 0's. So the default's become less reasonable, but we produce less garbage on updates. I think it's worthwhile.

Assets/Scripts/Utilities/RenderSettingsManager.cs
randomPoison, line 65: It has to be marked as `[SerializeField]` or `public` for `[ReadOnly]` to work. We should probably see if we can change `[ReadOnly]` to allow non-serialized members to still show up.

Assets/Scripts/Utilities/RenderSettingsManager.cs
randomPoison: I can do you one better:

```
float timeOfDay = timeOfDay = DayCycleManager.currentTime / ( DayCycleManager.dayCycleLength * ( 1.0f - _nightFraction) );
timeOfDay = Mathf.Clamp01( timeOfDay );
daylightIntensity = Mathf.Cos( timeOfDay * 2.0f * Mathf.PI ) * -0.5f + 0.5f;
```

PR #67, opened 2015-03-01
Sprint jump

brickman1444:
Makes jump take speed into account
Adds extension methods for vectors.


randomPoison:
How noticeable is the difference this makes? When I tried running and jumping (without these changes) I couldn't tell there was a speed difference.

PR #68, opened 2015-03-01
Variable jump height

brickman1444:
Hold jump to continue going up at a constant speed for a short amount of time. Releasing jump stops this. The effect is a variable height jump.

### DIFF
--- a/Assets/Scripts/Actors/ActorPhysics.cs
+++ b/Assets/Scripts/Actors/ActorPhysics.cs
@@ -79,7 +79,8 @@ public sealed class ActorPhysics : ActorComponent
 	 * Used for things like gliding or rolling where
 	 * The actor needs to keep moving in the last direction.
 	 */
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Actors/ActorPhysics.cs
```diff
@@ -107,8 +108,15 @@ public float normalizedMoveSpeed
...
 	[SerializeField] float jumpForce = 8.5f;
+	[Tooltip("Maximum length of time in seconds that the physics will push the player up while holding the jump button")]
+	[SerializeField] float _maxJumpForceTime = 0f;
+	public float maxJumpForceTime
+	{
+		get { return _maxJumpForceTime; }
+	}
 
 	[SerializeField] float _jumpCheckDistance = 1.0f;
+	[Tooltip("Keep this under the radius of the bumper")]
```
randomPoison, line 113: Add an assert to make sure this is the case.

PR #71, opened 2015-03-05
Showbuild - March 3

randomPoison:


Assets/Scripts/Actors/ActorPhysics.cs
```diff
@@ -58,7 +58,11 @@ public float sprintMoveSpeed
 		get { return _sprintMoveSpeed; }
 	}
 
-	float _jumpMoveSpeed = 0f;
+	[ReadOnly("Jump move speed")]
+	public float jumpMoveSpeed = 0f;
```
randomPoison: This should be `[SerializeField]`, not public since we don't want other scripts writing to it.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -82,7 +83,7 @@ public void GiveResource( PlayerStats actorStats, ResourceData resourceData)
 
 		_resources++;
 
-		actorStats.IncrementMaxStat( statType );
+		actorStats.SetMaxStat( statType, _resources * _statPerResource );
```
randomPoison: This should be a call to `RecalculateStatValue()`.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -95,7 +96,7 @@ public void DecrementResources()
 
 		if ( !_disableStatDecrease )
 		{
-			_ownerStats.DecrementMaxStat( statType );
+			_ownerStats.SetMaxStat( statType, _resources * _statPerResource );
```
randomPoison: This should be a call to `CalculateStatValue()`.

Assets/Scripts/Player/SkillFaker.cs
```diff
@@ -18,7 +18,7 @@ void Update()
 	{
 		if ( Input.GetKeyDown( _triggerKey ) )
 		{
-			_actorStats.IncrementMaxStat( _statToIncrease );
+			_actorStats.SetMaxStat( _statToIncrease, _actorStats.GetStatMaxValue( _statToIncrease ) + 0.1f );
```
randomPoison: Make the amount to increase the stat a variable.

randomPoison:
I'm done looking this over.

Assets/Scripts/Actors/ActorPhysics.cs
randomPoison: Nevermind, this value is set externally. I think we should instead look at a better way of managing the move speed when jumping. Rather than having a fixed move speed like that's set externally, we should have a max move speed (which would just be the sprinting speed), and just cap the move speed if the player keeps trying to move on one direction.

randomPoison: Fixed.

Assets/Scripts/Buddy/BuddyStats.cs
randomPoison: Fixed.

Assets/Scripts/Buddy/BuddyStats.cs
randomPoison: Fixed.

Assets/Scripts/Player/SkillFaker.cs
randomPoison: Done.

Assets/Scripts/Actors/ActorPhysics.cs
```diff
@@ -343,7 +343,13 @@ public void AirMovement( Vector3 inputVec, bool forceUp = false )
...
 		else
 		{
-			_moveVec = inputVec * _jumpMoveSpeed;
+			_moveVec += inputVec * _jumpMoveSpeedChangeRate;
+			_moveVec.y = 0.0f;
+
+			if ( _moveVec.sqrMagnitude > _sprintMoveSpeed * _sprintMoveSpeed )
+			{
+				_moveVec = _moveVec.normalized * _sprintMoveSpeed;
+			}
```
brickman1444, line 352: This might be a game feel thing but why isn't this being normalized to the sprint speed or grounded speed based on button input?

Assets/Scripts/BuddyManager.cs
```diff
@@ -23,6 +23,6 @@ public static void DecrementAllBuddyResources()
 			buddyStat.DecrementResources();
 		}
 
-		instance._buddyStats.RemoveAll( buddyStat => buddyStat.isAlive == false ); // Predicate statement
+		instance._buddyStats.RemoveAll( buddyStat => buddyStat.isAlive == false );
```
brickman1444, line 26: I'd recommend keeping this comment in here because I'm not going to remember what that => symbol means the next time I look at this and I won't even know what it's called.

Assets/Scripts/Nodes/BehaviorTreeInfo.cs
```diff
@@ -9,4 +9,9 @@ public class BehaviorTreeInfo : MonoBehaviour
 
 	public float moveSpeed;
 	public float watchDistance;
+
+	void OnDrawGizmos()
+	{
+		Gizmos.DrawWireSphere(transform.position, watchDistance);
```
brickman1444: pad with spaces

Assets/Scripts/Nodes/BuddyNodes.cs
```diff
@@ -47,6 +49,7 @@ public override NodeStatus Tick()
 		offset.z = offset.y;
 		offset.y = 0;
 		baseInfo.destination = transform.position + offset;
+		controller.moveDirection = Vector3.zero;
```
brickman1444: I'm actually not sure if this is necessary. I think it is but I don't know for certain.

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -179,6 +180,7 @@ public override void Update()
 			else
 			{
 				if ( player.controls.holdButton &&
+				     player.stats.CanUseStat( Stat.Stamina ) &&
```
brickman1444, line 183: Was this really it to fix that climbing bug?

Assets/Scripts/Player/PlayerStats.cs
```diff
@@ -50,23 +49,18 @@ public override void Awake()
...
-		statObject.currentMax += statObject.maxIncrement;
-		ScaleCurrImage( statObject );
-	}
+		DebugUtils.Assert( maxValue > 0.0f, "Max stat value must be greater than 0." );
 
-	public void DecrementMaxStat( Stat stat )
-	{
 		StatObject statObject = _statDictionary[stat];
-		statObject.currentMax = Mathf.Max(statObject.currentMax - statObject.maxIncrement, 0.0f); // decrement and clamp at a minimum of 0
+		statObject.currentMax = maxValue; // assume maxValue is a sane value.
```
brickman1444: I don't think you really need this comment here. The assert should be pretty explanatory

Assets/Scripts/Utilities/MathUtils.cs
```diff
@@ -47,4 +48,36 @@ public static Vector3 SetZ( this Vector3 vec, float z )
...
 	}
+	#endregion
+
+	#region Colliders
+
+	public static bool IsWithinInfiniteVerticalCylinders( Vector3 testPoint, CapsuleCollider[] colliders )
+	{
+		foreach ( CapsuleCollider collider in colliders )
+		{
+			if (IsWithinInfiniteVerticalCylinder( testPoint, collider ) )
```
brickman1444: missing space at the front here

brickman1444:
Done. Good job putting this together

Assets/Scripts/Actors/ActorPhysics.cs
randomPoison, line 352: The horizontal move speed while in the air shouldn't be a constant value, that was the problem with your implementation. If the player is in the air and moving forward and starts holding the back button, they should slow down to a stop (horizontally) before starting to move backwards. So if we normalize the movement speed then there's no in-between, they always move at a set speed in whatever direction they're pressing.

Assets/Scripts/BuddyManager.cs
randomPoison, line 26: That's a language feature, though. We shouldn't be putting comments in every time someone doesn't understand some aspect of the language. If you see it again and forget what it does, you can look up `RemoveAll()` and see that it takes a predicate statement as an argument.

Assets/Scripts/Nodes/BehaviorTreeInfo.cs
randomPoison: Done.

Assets/Scripts/Nodes/BuddyNodes.cs
randomPoison: It shouldn't be there. Since the node isn't related to movement, it shouldn't be affecting movement. If we want to be sure that the buddy stops moving when the movement node ends, the movement node should be doing taking care of that.

Assets/Scripts/Player/PlayerControls.cs
randomPoison, line 183: Yup, I just wasn't checking if the player had enough stamina before trying to climb.

Assets/Scripts/Player/PlayerStats.cs
randomPoison: Done.

PR #73, opened 2015-03-07
Teleporter

brickman1444:
Teleporter prefab with target location as a child object. Sound plays on teleport.


Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -490,12 +490,12 @@ public void Respawn()
 		SoundManager.Play2DSound( _respawnSound );
 	}
 
-	private void Teleport( Vector3 toPosition, Quaternion toRotation = new Quaternion() )
+	public void Teleport( Vector3 toPosition, Quaternion toRotation = new Quaternion() )
 	{
-		transform.position = _respawnPosition;
-		transform.rotation = _respawnRotation;
+		transform.position = toPosition;
+		transform.rotation = toRotation;
```
brickman1444, line 496: This was a typo from before. It is fixed now.

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -490,12 +490,12 @@ public void Respawn()
 		SoundManager.Play2DSound( _respawnSound );
 	}
 
-	private void Teleport( Vector3 toPosition, Quaternion toRotation = new Quaternion() )
+	public void Teleport( Vector3 toPosition, Quaternion toRotation = new Quaternion() )
```
randomPoison: Instead of `new Quaternion()` do `Quaternion.identity`.

Assets/Teleporter.cs
```diff
@@ -0,0 +1,37 @@
...
+
+	void OnTriggerEnter( Collider col )
+	{
+		if ( col.gameObject.transform.parent == null )
+		{
+			return;
+		}
+
+		// Get to the player from the bumper
+		PlayerControls controls = col.gameObject.GetComponent<Transform>().parent.gameObject.GetComponent<PlayerControls>();
```
randomPoison: You could use [`GetComponentInParent()`](http://docs.unity3d.com/ScriptReference/GameObject.GetComponentsInParent.html) instead of hunting down the transform.

Assets/Teleporter.cs
```diff
@@ -0,0 +1,37 @@
...
+
+		if ( controls != null )
```
randomPoison: You can just do `if ( controls )`, which is more consistent with what we've been doing elsewhere.

randomPoison:
Try not teleporting the camera with the player. It will automatically fly over to the player's new position, but it might help to make it not feel as jarring since the motion of the camera wont' be as instant. Later on we'll want to add a path for the camera to fly along so that it's a smoother trip and doesn't fly through geometry.

brickman1444:
Closing to reopen on master

PR #74, opened 2015-03-07
Controls screen

brickman1444:
Adds controls screen with "press space to start" functionality. 
Adds a first pass at a level abstraction. Typing in level names by string into the inspector felt really bad to me. I think an even better way could be a serializable object with a dictionary on it that can be editted in the inspector. That would take out the remaining hardcoded level names in code.


randomPoison:
Rather than writing a script that handles input, do this with the Unity UI and event system. Also, is there a reason why `LoadGameOnButton.cs` is under `Assets/` and not `Assets/Scripts/`?

PR #80, opened 2015-03-09
Add alternate buttons for scrolling through inventory

brickman1444:
for people without scroll wheels, like Zac

### DIFF
--- a/Assets/Scripts/Player/PlayerInventory.cs
+++ b/Assets/Scripts/Player/PlayerInventory.cs
@@ -74,7 +74,7 @@ public class PlayerInventory : ActorComponent
 
 		float scrollAmount = InputUtils.GetAxis("Scroll");
 
-		if ( scrollAmount < WadeUtils.SMALLNUMBER && scrollAmount > -WadeUtils.SMALLNUMBER )
+		if ( WadeUtils.IsZero( scrollAmount ) )
 		{
 			if ( _altScrollButton.positiveDown )
 			{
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Player/PlayerInventory.cs
```diff
@@ -58,19 +60,39 @@ void Start()
...
-			int nextIndex = resourceIndex + Mathf.Clamp( Mathf.RoundToInt( scrollAmount ), -1, 1 );
-			resourceIndex = MathUtils.Mod( nextIndex, heldResources.Count );
-			SpawnResourceObject();
+			// Nothing to scroll
+			return;
+		}
+
+		float scrollAmount = InputUtils.GetAxis("Scroll");
+
+		if ( scrollAmount < WadeUtils.SMALLNUMBER && scrollAmount > -WadeUtils.SMALLNUMBER )
```
randomPoison: I think there's an `IsZero()` method for floats in `MathUtils`. If not, now would be a good time to add one.

Assets/Scripts/Player/PlayerInventory.cs
```diff
@@ -58,19 +60,39 @@ void Start()
...
+			{
+				scrollAmount = 1f;
+			}
+			else if ( _altScrollButton.negativeDown )
+			{
+				scrollAmount = -1f;
+			}
+			else
+			{
+				return;
```
randomPoison: The control flow here isn't very clear, it took me a long time to understand what's going on here. Get rid of the early return, it's not necessary and only makes the method harder to read.

randomPoison:
I'm done looking this one over.

Assets/Scripts/Player/PlayerInventory.cs
```diff
@@ -58,16 +60,35 @@ void Start()
...
+				scrollAmount = -1f;
+			}
+		}
+
+		if ( !WadeUtils.IsZero( scrollAmount ) )
+		{
+			int nextIndex = resourceIndex + ( scrollAmount < 0f ? -1 : 1 );
 			resourceIndex = MathUtils.Mod( nextIndex, heldResources.Count );
 			SpawnResourceObject();
 		}
```
brickman1444, line 94: How is this for readability?

excaliburHisSheath: Use an `else` statement so you're not doing two conditional branches.

randomPoison, line 94: Add a comment on that second block to explain why you need to check for zero a second time.

PR #84, opened 2015-03-14
Planet VFX

randomPoison:
- Planet now looks like Earth.
- Clouds now move over surface of planet.
- Planet has a rotating dark side.
- Planet rotates slowly.
- The dark side of the planet is transparent.
- The level of the transparency changes with the light level, so that it's more transparent during the day.


Assets/Scripts/VFX/PlanetLight.cs
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	void Update()
+	{
+		// rotate light direction
+		Quaternion rotation = Quaternion.Euler( 0.0f, _rotationSpeed * Time.deltaTime, 0.0f );
+		_lightDirection = rotation * _lightDirection;
+		_renderer.material.SetVector(
+			"_LightDir",
+			new Vector4( _lightDirection.x, _lightDirection.y, _lightDirection.z ) );
```
brickman1444: Why is this a new vector instead of just _lightDirection?

brickman1444:
Done. I'm not that familiar with shaders though. @cjacobwade should look at this

Assets/Scripts/VFX/PlanetLight.cs
randomPoison: `SetVector()` takes a `Vector4`, I wanted to use a Vector3 for `_lightDirection` to ensure that it would rotate correctly when applying the quaternion, but it occurs to me now that a Vector4 isn't a 4D vector, it's a 3D vector with a homogeneous coordinate, so I'll fix that.

randomPoison: Actually, `_lightDirection` was already a Vector4, so I don't know what smoke I was cracking.

PR #85, opened 2015-03-16
Buddy Death and Resource Respawning

brickman1444:
Buddies fall on their face when they die and output a nearly permanent death emote.
Buddy eggs respawn when their associated buddy dies.
Regular resources respawn after a set amount of time.


Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -66,6 +66,18 @@ enum HappinessState
 	[SerializeField] float _emoteRoutineWait = 0f;
 	Coroutine _currentEmoteRoutine = null;
 
+	[Header( "Debug Settings" )]
+	[SerializeField] bool _disableStatDecrease = false;
```
randomPoison: Debug settings should be at the bottom. As it stands `particlesRenderer` and `_deadParticleSystem` are showing up under the "Debug Settings" header. Also, make sure you do `[Space( 10 )]` when you do `[Header()]`.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -66,6 +66,18 @@ enum HappinessState
...
+
+	[SerializeField] ParticleSystem _particles;
+	Renderer _particlesRenderer;
+	[SerializeField] ParticleSystem _deadParticleSystem;
```
randomPoison, line 71: See if it's possible to change the way you're doing this so we don't need a second particle system. It doesn't seem like we're gaining much from having a second one, and it only makes things more complicated.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -311,6 +327,18 @@ public void Kill()
 	{
 		_isAlive = false;
 		Destroy( GetComponent<AIController>() );
+		_happiness = 0;
+		RecalculateStat();
+		EmoteDeath();
+		actor.physics.ComeToStop(); // This doesn't quite work and I don't know why
```
randomPoison: `ComeToStop()` needs to be called every frame, it doesn't set any internal state for the physics system. You should set something up within the behavior physics controller to actually transition to a state where it just keeps the buddy in place.

Assets/Scripts/Managers/DayCycleManager.cs
```diff
@@ -48,6 +48,12 @@ void Start()
 		}
 	}
 
+	void OnValidate()
+	{
+		CancelInvoke( "StartMidnightOverlay" );
+		Invoke( "StartMidnightOverlay", ( _dayCycleLength - _blackoutDuration ) - _currentTime );
```
randomPoison: This calculation is done in multiple places, so that indicates it should probably be its own function.

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444: Fixed

randomPoison:
At this point `InventoryItem` should be split into two classes, one for resources and one for buddy eggs.

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444, line 71: I'm not sure if the particles would properly outlive the particle system, but I do know for certain that you can't change the color over time, which we would need to do. The emotes fade in and out but if the death did that it would look bad. So I don't think we can do this with one particle system.

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444: Fixed

Assets/Scripts/Managers/DayCycleManager.cs
brickman1444: Fixed

brickman1444:
Made an object hierarchy for InventoryItems, BuddyItems and Resource Items.

PR #86, opened 2015-03-16
Instant happiness

brickman1444:
Buddies now get an instant happiness boost when being fed and not overfed.
Buddies can only lose happiness overnight, not gain it.


randomPoison:
Why are a bunch of Chris's and Parker's commits showing up as part of this? Make sure it's just the relevant changes before I review this.

randomPoison:
Reopen this pull request once the buddy death stuff has been merged into master. Then only the relevant commits should show up.

PR #87, opened 2015-03-16
Resource holder polish

brickman1444:
Adds cool particle shader to resource holders
Adds particle system to resource holders
Scales down resource holder and particle system if under an overhang.


randomPoison:
Yeah, this seems fine to me.

PR #89, opened 2015-03-17
Checkpoint lifters & Rotating skybox

cjacobwade:
working (with somewhat hacky solution for activating w....../ buddy births), also added slow rotation to the skybox


Assets/Scripts/Managers/RenderSettingsManager.cs
```diff
@@ -79,10 +81,13 @@ public static float daylightIntensity
 	void Awake()
 	{
 		_curSkyboxTintPropertyID = Shader.PropertyToID( "_Tint" );
+		RenderSettings.skybox.SetFloat( "_Rotation", 0f );
 	}
 
 	void Update()
 	{
+		RenderSettings.skybox.SetFloat( "_Rotation", Time.realtimeSinceStartup * _skyboxRotSpeed );
```
brickman1444, line 90: is this intended to be associated with the time of day at all or does it just spin on its own?

Assets/Scripts/Managers/BuddyManager.cs
```diff
@@ -14,6 +17,17 @@ void Start()
 	public static void RegisterBuddy( BuddyStats buddyStats )
 	{
 		instance._buddyStats.Add( buddyStats );
+
+		instance.ActivateCheckpoint();
+	}
+
+	void ActivateCheckpoint()
+	{
+		if( checkpointIndex < checkpointLifters.Length && checkpointLifters[checkpointIndex] != null )
```
brickman1444: I think the condition after the && here could just be "checkpointLifters[checkpointIndex]"

Assets/Scripts/Managers/RenderSettingsManager.cs
```diff
@@ -79,10 +81,13 @@ public static float daylightIntensity
 	void Awake()
 	{
 		_curSkyboxTintPropertyID = Shader.PropertyToID( "_Tint" );
+		RenderSettings.skybox.SetFloat( "_Rotation", 0f );
```
brickman1444: consider storing this string as a field

Assets/Scripts/Resources/CheckpointLifter.cs
```diff
@@ -0,0 +1,71 @@
...
+{
+	[SerializeField] float moveSpeed = 5f;
+
+	Vector3 initPos = Vector3.zero;
+	[SerializeField] float maxHeightOffset = 25f;
+
+	bool isActive = false;
+	bool isRising = false;
+
+	Renderer rend;
```
brickman1444: type out _renderer.

Assets/Scripts/Managers/RenderSettingsManager.cs
cjacobwade, line 90: This is just on its own. It's just clouds spinning around the y-axis.

Assets/Scripts/Resources/CheckpointLifter.cs
```diff
@@ -0,0 +1,71 @@
...
+				}
+			}
+			else if( transform.position.y > initPos.y )
+			{
+				transform.position -= Vector3.up * moveSpeed * Time.deltaTime;
+			}
+			else
+			{
+				transform.position = initPos;
+			}
```
brickman1444: I'd make this whole block of if ( isRising ) { blah } else { blah } symmetrical for readability so you can easily tell that both sides are doing the same thing but opposites.

Assets/Scripts/Resources/CheckpointLifter.cs
```diff
@@ -0,0 +1,71 @@
...
+			isRising = true;
+		}
+	}
+
+	void OnTriggerExit( Collider otherCol )
+	{
+		if ( otherCol.GetComponentInParent<PlayerActor>() )
+		{
+			isRising = false;
+		}
```
brickman1444: so if the player jumps at all when the lifter is rising it immediately starts to go down? a short delay would be nice.

Assets/Scripts/Managers/BuddyManager.cs
```diff
@@ -6,6 +6,9 @@ public class BuddyManager : SingletonBehaviour<BuddyManager> {
 
 	List<BuddyStats> _buddyStats = new List<BuddyStats>();
 
+	[SerializeField] CheckpointLifter[] checkpointLifters = new CheckpointLifter[1]; // Need these to activate checkpoints
```
brickman1444: why does this start with size 1? does that even matter?

Assets/Scripts/Managers/BuddyManager.cs
```diff
@@ -6,6 +6,9 @@ public class BuddyManager : SingletonBehaviour<BuddyManager> {
...
+	int checkpointIndex = 0; // Index of next checkpoint to activate
```
brickman1444: consider making this ReadOnly so we can see what lifter we're on while debugging

Assets/Scripts/Resources/CheckpointLifter.cs
```diff
@@ -0,0 +1,71 @@
+﻿using UnityEngine;
+using System.Collections;
+
+public class CheckpointLifter : MonoBehaviour 
+{
+	[SerializeField] float moveSpeed = 5f;
+
+	Vector3 initPos = Vector3.zero;
```
brickman1444: This could be ReadOnly if you think that'd be helpful for debugging

Assets/Scripts/Resources/CheckpointLifter.cs
cjacobwade: done

Assets/Scripts/Managers/RenderSettingsManager.cs
cjacobwade: done

Assets/Scripts/Managers/RenderSettingsManager.cs
brickman1444, line 90: kk

Assets/Scripts/Managers/BuddyManager.cs
cjacobwade: changed

brickman1444:
Done. Adding a pause before it starts to descend would be nice. Another even better improvement would be to have a child object that is the target and have the lifter move toward that object. This would allow for horizontal and diagonal movement and be easy to set up in the scene editor. This is probably out of scope for this pull request. EDIT: oo and drawing a debug line from the initpos to the target pos would look so good.

Assets/Scripts/Resources/CheckpointLifter.cs
cjacobwade: done

Assets/Scripts/Resources/CheckpointLifter.cs
cjacobwade: I'm using a trigger zone that's higher than the ground by a bit so the short delay happens anyway

Assets/Scripts/Managers/BuddyManager.cs
cjacobwade: Need to manually hook these up so they can't be readonly

brickman1444: The index could be readonly couldn't it?

Assets/Scripts/Resources/CheckpointLifter.cs
brickman1444: cool

Assets/Scripts/Managers/BuddyManager.cs
cjacobwade: oh yeah, fixed

Assets/Scripts/Resources/CheckpointLifter.cs
cjacobwade: done

PR #90, opened 2015-03-17
Buddy walk is dependent on their happiness level

cjacobwade:
yeah

### DIFF
--- a/Assets/Scripts/Buddy/BuddyStats.cs
+++ b/Assets/Scripts/Buddy/BuddyStats.cs
@@ -212,6 +212,8 @@ public class BuddyStats : ActorComponent
 			_happiness = 0;
 		}
 
+		actor.animator.SetFloat( "happiness", _happiness );
+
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -212,6 +212,8 @@ public void AdjustHappiness( float deltaHappiness )
 			_happiness = 0;
 		}
 
+		actor.animator.SetFloat( "happiness", _happiness );
```
brickman1444, line 215: This assumes that it's normalized, right? Because that's not in yet. I can put it in tonight though.

brickman1444:
Everything looks fine.

Assets/Scripts/Buddy/BuddyStats.cs
cjacobwade, line 215: Yeah it does assume normalized. If it goes over one it won't cause errors, but the buddies will seem extra happy.

PR #91, opened 2015-03-17
Multi Buddy Support

brickman1444:
Gets stat off of aggregate happiness
Limits happiness to [0,1]
Adds function of number of buddies for calculating nightly resource drain.
Stops spawning buddies under the hill.


Assets/Editor/CustomAttributeDrawers.cs
```diff
@@ -19,7 +19,7 @@ public class ReadOnlyDrawer : PropertyDrawer
 		ReadOnlyAttribute readOnlyAttribute = (ReadOnlyAttribute)attribute;
 
 		// Use a specified name if there is one
-		string displayName = ( readOnlyAttribute.displayName != "" ) ? readOnlyAttribute.displayName : property.name;
+		string displayName = ( readOnlyAttribute.displayName != "" ) ? readOnlyAttribute.displayName : property.displayName;
```
brickman1444, line 22: The displayName is the "pretty name". I don't know why I didn't see it before. Maybe it's a Unity 5 thing. Maybe I'm dumb. Moral of the story is you don't have to put in a name for ReadOnly if the default name style works.

Assets/Prefabs/Actors/Buddy.prefab
```diff
@@ -1291,15 +1291,13 @@ MonoBehaviour:
...
-  _happinessIncrementPerResource: .150000006
+  _nightlyHappinessDecrement: .100000001
+  _happinessIncrementPerResource: .0299999993
   _neutralHappinessRange:
-    min: 12
-    max: 20
-  _startingHappiness: 10
+    min: .400000006
+    max: .699999988
+  _startingHappiness: .300000012
```
cjacobwade, line 1299: is this a midi keyboard number? or does the curve have to work this way?

brickman1444, line 1299: I think that's a floating point thing. I put in sensible numbers

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -204,14 +206,13 @@ public void AdjustHappiness( float deltaHappiness )
...
 			_happiness = 0;
 		}
-
-		RecalculateStat();
-
-		if ( _happiness < 0 )
+		else if ( _happiness > 1 )
 		{
-			_happiness = 0;
+			_happiness = 1;
```
cjacobwade: Should use 1f for this and similar situations. It's a miniscule performance boost and a good habit to get into

Assets/Scripts/Player/PlayerInventory.cs
```diff
@@ -164,9 +164,24 @@ void SpawnBuddy()
...
+		else
+		{
+			spawnLocation = transform.position + transform.forward * _buddySpawnDistance;
+		}
+
 		BuddyItemData buddyItemData = (BuddyItemData)heldResources[resourceIndex];
 		BuddyStats newBuddy = ( Instantiate( buddyItemData.buddyPrefab,
-		                                     transform.position + transform.forward * _buddySpawnDistance,
+		                                     spawnLocation,
 		                                     Quaternion.identity ) as GameObject ).GetComponent<BuddyStats>();
```
cjacobwade, line 185: Would be nice to have the buddy spawn looking at the player (and with an effect), but that's a polish thing.

Assets/Scripts/Buddy/BuddyStats.cs
brickman1444: This seems like something that would be compiled out to me. This guy agrees http://stackoverflow.com/a/16170339

Assets/Scripts/Utilities/CustomClassDrawers.cs
```diff
@@ -0,0 +1,76 @@
...
+
+[CustomPropertyDrawer( typeof( BuddyResourceCurve ) )]
+public class BuddyResourceCurveDrawer : PropertyDrawer
+{
+	float textHeight = 15f;
+
+	public override float GetPropertyHeight( SerializedProperty property,
+	                                         GUIContent label )
+	{
+		return textHeight * 11;
```
cjacobwade: Magic number?

Assets/Scripts/Buddy/BuddyStats.cs
cjacobwade: I tested on my own and got an average of 6ms faster with f, but maybe my test was flawed...Or that guy could be on a newer version of .NET. It doesn't matter really

PR #92, opened 2015-03-17
Fix Buddy Performance

randomPoison:
- Changed buddy behavior tree to not EAT ALL THE FRAME TIME.


randomPoison:
Since it's just an asset change, I'm going to merge this in. It's a minor fix, anyway. The more major behavior tree stuff is coming down the pipe.

brickman1444:
Looks good to me

PR #94, opened 2015-03-20
Fix Resource Holders

brickman1444:
- Resource holders no longer stick up through the ground.
- Fixed a bug with the buddy behavior tree that was causing errors.


Assets/Scripts/Resources/ResourceHolder.cs
```diff
@@ -16,21 +19,23 @@ public class ResourceHolder : MonoBehaviour
...
-		_beaconRenderer = GetComponentInChildren<BeaconTag>().gameObject.GetComponent<Renderer>();
+		_beaconRenderer = _beacon.GetComponent<Renderer>();
 		_particleSystem = GetComponentInChildren<ParticleSystem>();
 
 		RaycastHit hitInfo;
 
-		Physics.Raycast( new Ray( _transform.position, Vector3.up), out hitInfo, _maxHeight );
+		Physics.Raycast( new Ray( _transform.position, Vector3.up ), out hitInfo, _maxHeight, _collisionLayer );
+
+		float normalizedHeight = _maxHeight;
```
brickman1444: Normalized height sounds like it should be from 0 to 1. Is that what this is?

Assets/Scripts/Resources/ResourceHolder.cs
```diff
@@ -16,21 +19,23 @@ public class ResourceHolder : MonoBehaviour
...
 		{
-			// The .5 is to correct for some scale thing
-			float normalizedHeight = ( hitInfo.transform.position.y - _transform.position.y ) / ( _maxHeight * 0.5f );
-			_transform.localScale = _transform.localScale.SetY( normalizedHeight );
-			_particleSystem.startLifetime *= normalizedHeight;
+			normalizedHeight = hitInfo.transform.position.y - _transform.position.y;
 		}
 
+		_beacon.localScale = _beacon.localScale.SetY( normalizedHeight );
+		_particleSystem.startLifetime = normalizedHeight / _particleSystem.startSpeed * 0.5f; // when the speed of a particle is 1
```
brickman1444: Does this depend on the speed? I.e if we speed up the particles but shorten the lifetime accordingly, does this line have to change?

The only reason I ask is because the comment makes me think that.

randomPoison: Whoops, I didn't finish that comment. The point of this line is to adjust the lifetime of the particle based on its speed so that it stays within the beacon (and so doesn't poke through the ceiling). The hardcoded `0.5f` is because it seems like the particles actually move twice as fast as the `startSpeed` value says. I haven't looked further into this since we're short on time, so this is only a quick solution.

Assets/Scripts/Resources/ResourceHolder.cs
randomPoison: Nope, I should rename that variable.

PR #95, opened 2015-04-13
Text pop ups

brickman1444:
Text pop ups are available in two ways: tagging an object with and Interactable component makes it trigger a timed text pop up on click. Creating a trigger volume and tagging it with a TextVolume component makes a text pop up appear on enter and clear on exit. All text pop ups can be cancelled with a button.


cjacobwade:
Can you put these scripts into some folder besides assets and add the needed stuff to a scene so we can try this out? I tried creating a volume with that tag, but it looks like there's stuff missing from the player prefab and scene.

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -403,16 +424,30 @@ public override void Exit()
...
 	}
 	#endregion
 
-	/**
+	public bool InteractCheck( RaycastHit hitInfo )
+	{
+		Interactable interactable = hitInfo.collider.gameObject.GetComponent<Interactable>();
+		if ( interactable )
+		{
+			_textBox.SetTextForDuration( interactable.interactText, interactable.duration );
```
cjacobwade, line 441: Might make more sense to move this out into a DoInteract method where we can call other callbacks as well

Assets/Scripts/UI/TextBox.cs
```diff
@@ -0,0 +1,50 @@
...
+public class TextBox : MonoBehaviour {
+
+	Text _UIText = null;
+
+	[SerializeField] KeyCode _clearKey = 0;
+
+	void Start ()
+	{
+		_UIText = GetComponent<Text>();
+	}
```
cjacobwade, line 14: Could probably just inherit from Text instead of holding a reference to another component.

randomPoison:
This commit is now part of #98.

PR #96, opened 2015-04-13
Buddy Actions

randomPoison:
- Made some major updates to the behavior tree library.
  - All tree nodes are in separate files.
  - Had to replace all existing behavior trees since the new format is not backwards compatible.
- Introduced framework for buddy idle actions. Buddies will wander around as before, but once their happiness is about 0.9 they will occasionally do a "hug" action (this will be replaced with "dance" once the animation is made).
- Improved buddy behavior when following player: They now stop before running into the player, and they go back to idling when they are no longer hungry.


Assets/Scripts/Nodes/Buddy Nodes/CheckHunger.cs
```diff
@@ -0,0 +1,43 @@
+﻿using UnityEngine;
+using System.Collections;
+using BehaviorTree;
+
+public enum Comparison
+{
+	GreaterThan,
+	EqualTo,
+	LessThan
+}
```
cjacobwade: Probably should move this out somewhere like BuddyBehaviourUtils just so its not in a random node

Assets/Scripts/Nodes/FindGod.cs
```diff
@@ -0,0 +1,27 @@
...
+	private Hashtable _data;
+
+	public override void InitSelf( Hashtable data )
+	{
+		_data = data;
+	}
+
+	public override NodeStatus TickSelf()
+	{
+		GodTag targetGod = GameObject.FindObjectOfType<GodTag>();
```
cjacobwade, line 17: Does this happen every frame? If so can we move this into InitSelf and have a callback that updates this when needed? Really skeptical of FindObject each frame.

Assets/Scripts/Nodes/FollowTarget.cs
```diff
@@ -0,0 +1,36 @@
...
+	{
+		GameObject gameObject = (GameObject)data["gameObject"];
+		_transform = gameObject.GetComponent<Transform>();
+		_controller = gameObject.GetComponent<BehaviorPhysicsController>();
+		_target = (Transform)data["target"];
+	}
+
+	public override NodeStatus TickSelf()
+	{
+		_controller.moveDirection = ( _target.position - _transform.position ).normalized;
```
cjacobwade, line 24: Do we need to set y value of this to zero to keep buddies from trying to fly?

Assets/Scripts/Nodes/FollowTarget.cs
```diff
@@ -0,0 +1,36 @@
...
+
+		if ( Vector3.Distance( _transform.position, _target.position ) < minDistance )
```
cjacobwade: Could change this to use square magnitude and squared minDistance since distance does sqrt

Assets/Scripts/Nodes/MoveToDestination.cs
```diff
@@ -0,0 +1,38 @@
...
+		_destination = (Vector3)data["destination"];
+	}
+
+	public override NodeStatus TickSelf()
+	{
+		_controller.moveDirection = ( _destination - _transform.position ).normalized;
+
+		Debug.DrawLine( _transform.position, _destination );
+
+		if ( Vector3.Distance( _transform.position, _destination ) < minDistance )
```
cjacobwade: Could do squared minDistance and magnitude

cjacobwade:
Done looking things over.

Assets/Scripts/Nodes/Buddy Nodes/CheckHunger.cs
randomPoison: Done.

Assets/Scripts/Nodes/FindGod.cs
randomPoison, line 17: The node is fine, you just have to setup the tree so this only get's called once at the beginning (the same way you'd only call `FindObject()` and cache the value).

Assets/Scripts/Nodes/FollowTarget.cs
randomPoison, line 24: No, the `PlayerPhysics` system ensures that won't happen.

Assets/Scripts/Nodes/FollowTarget.cs
randomPoison: Done.

Assets/Scripts/Nodes/MoveToDestination.cs
randomPoison: Done.

PR #97, opened 2015-04-16
Shader utils

cjacobwade:


Assets/Scripts/Utilities/MassAdjustOcclusionBounds.cs
```diff
@@ -0,0 +1,27 @@
...
+//				bounds.center = mesh.bounds.center;
+//				bounds.size = mesh.bounds.size * 2f;
+//
+//				mesh.bounds = bounds;
+//				mf.mesh = mesh;
+//			}
+//		}
+//
+//		Debug.Log( counter );
+//	}
```
randomPoison: You can just delete the commented stuff.

PR #98, opened 2015-04-18
Tutorial

randomPoison:
- Adds a tutorial section to the level.
- Fixes some of the code for displaying text boxes.
- Cleans up the level some and removes some unneeded files.


Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -403,16 +420,30 @@ public override void Exit()
...
+		_actor.physics.RegisterState( PhysicsStateType.Climbing,  new Climbing( _actor ) );
+		_actor.physics.RegisterState( PhysicsStateType.Gliding,   new Gliding( _actor ) );
+		_actor.physics.RegisterState( PhysicsStateType.Sprinting, new Sprinting( _actor ) );
 	}
 	#endregion
 
-	/**
+	public bool InteractCheck( RaycastHit hitInfo )
+	{
+		Interactable interactable = hitInfo.collider.gameObject.GetComponent<Interactable>();
```
cjacobwade, line 434: Can you rename this class to something more descriptive, like TextInteractable or TextPopup? Generic is good, but we only have one use case for this planned.

Assets/Scripts/Player/PlayerInventory.cs
```diff
@@ -189,6 +198,7 @@ void SpawnBuddy()
 
 		inventory[heldResources[resourceIndex]]--;
 		UpdateResourceList();
+		return true;
```
cjacobwade, line 201: I feel like this function would make more sense logically if this ordering was reversed. if(WithinCylinder) {SpawnBuddy; return true; } else { return false; }

Assets/Scripts/Resources/ResourceHolder.cs
```diff
@@ -7,10 +7,10 @@ public class ResourceHolder : MonoBehaviour
 
 	[SerializeField] float resourceHeightOffset = 0.17f;
```
cjacobwade: Can you make this _resourceHeightOffset?

Assets/Scripts/TextVolume.cs
```diff
@@ -0,0 +1,56 @@
...
+	void Awake()
+	{
+		_textBox = FindObjectOfType<TextBox>();
+	}
+
+	void OnDrawGizmos()
+	{
+		Gizmos.color = Color.blue;
+		Transform transform = GetComponent<Transform>();
+		Gizmos.matrix = Matrix4x4.TRS( transform.position, transform.rotation, transform.lossyScale );
```
cjacobwade: I think there's a shorthand for this, something like transform.localToWorldMatrix

Assets/Scripts/UI/TextBox.cs
```diff
@@ -0,0 +1,70 @@
...
+	{
+		_UIText = GetComponent<Text>();
+	}
+
+	void Update()
+	{
+		if ( Input.GetKeyDown( _clearKey ) )
+		{
+			ClearText();
+			CancelInvoke( "ClearText" );
```
cjacobwade: Where is this being Invoked that it needs to be cancelled? I would expect that to be in a public function on this object.

Assets/Scripts/UI/TextBox.cs
```diff
@@ -0,0 +1,70 @@
...
+			yield return null;
+
+			duration -= Time.deltaTime;
+		}
+		while ( duration > 0.0f );
+
+		// Clear the text.
+		_UIText.text = "";
+		_textFade = null;
+	}
```
cjacobwade, line 67: This totally isn't a fade... Maybe you meant to add an alpha lerp towards zero?

Assets/Scripts/UI/TextBox.cs
```diff
@@ -0,0 +1,70 @@
...
+
+	public void ClearText()
+	{
+		_UIText.text = "";
+
+		if ( _textFade != null )
+		{
+			StopCoroutine( _textFade );
+		}
```
cjacobwade, line 48: Do you need to null out _textFade here or is that automatic?

Assets/Scripts/Player/PlayerControls.cs
randomPoison, line 434: I don't think that's a good idea. A better approach would be to unify cutting and clicking on buddies into a single "interactible" system. I'll add that to the GitHub issues.

cjacobwade:
I'm not seeing an option to stop the player from moving anywhere in here. I agree that the movement tutorial shouldn't have that, but it makes sense for other text popups to interrupt player movement.

Assets/Scripts/Player/PlayerInventory.cs
randomPoison, line 201: I'd disagree only because that would put the entire body of the function within the if statement, whereas the way it is now keeps things a bit cleaner.

Assets/Scripts/Resources/ResourceHolder.cs
randomPoison: Done.

Assets/Scripts/TextVolume.cs
randomPoison: Cool, thanks.

Assets/Scripts/UI/TextBox.cs
randomPoison: Whoops, that's no longer necessary. Fixed.

randomPoison:
Other text popups aren't in yet, this PR is only for the tutorial.

Assets/Scripts/UI/TextBox.cs
randomPoison, line 67: Yeah, I'll add that in later. I'm rushing this PR through because nobody else can work in the scene until it's merged in.

Assets/Scripts/UI/TextBox.cs
randomPoison, line 48: I'm guessing it's automatic, though on that note I should remove the line in the coroutine that does null it out.

PR #100, opened 2015-04-18
Text Box Fade Out

randomPoison:
- Adds a panel behind the tutorial text popups to make them easier to read.
- Adds text fadeout.


Assets/Scripts/Utilities/MathUtils.cs
```diff
@@ -85,4 +85,10 @@ public static bool RandBool()
 	}
 
 	#endregion
+
+	public static Color SetAlpha( this Color color, float alpha )
```
cjacobwade, line 89: I think a cleaner way to do this would be ColorUtils.SetAlpha( ref Color color, float alpha). Then using it in code doesn't require the weird readability of color = color.SetAlpha(blah)

randomPoison, line 89: `this` already passes by reference, the problem is the property used to get at the color in the first place passes the `Color` by value, so you still have to do the assignment.

```
_UIText.color = _UIText.color.SetAlpha( 1.0f );
```

To clarify, if I just did `_UIText.color.SetAlpha( 1.0f )` it wouldn't work because `_UIText.color` is a property and returns a copy of the `Color` object, so I have to do the assignment to actually change `_UIText.color`.

PR #108, opened 2015-05-02
Step Sounds and Ambient Audio

brickman1444:
Scene Changes required:
- Replace the main tree with the Main Tree Prefab from the Prefabs/Environment/Garden/Trees folder so that it has the ambient spring sounds audio source attached.
- Replace the pond with the Pond Prefab from the Prefabs/Environment/Garden folder so that it has a water volume attached
- Place trigger volumes with the Brush Volume component attached in any areas where the player's steps should sound like they are walking through brush. I don't know if we need this.


Assets/Scripts/Actors/TravelSoundPlayer.cs
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	public AudioClip GetRandomClip()
+	{
+		return _travelSounds[Random.Range( 0, _travelSounds.Length )];
+	}
+
+#if UNITY_EDITOR
+	[MenuItem( "Assets/Create/Travel Sound Player" )]
+	public static void CreateTravelSoundPlayerAsset()
```
randomPoison: A better place for this method would be in `Assets/Editor/MenuItems.cs`. That's where I've been putting all the other menu item methods like this, and it would mean you don't need to wrap stuff in `#if UNITY_EDITOR`.

randomPoison:
Done with my review.

brickman1444:
The scene changes still have to get done.

PR #109, opened 2015-05-02
Adult Buddy Fixes

randomPoison:
- Fixes the issue where adult buddies would reappear after midnight.
- Fixes the issue the stats gained from an adult buddy would go away after midnight.
- Fixes a bug in SoundManager that would cause errors if an audio source is inadvertently destroyed.
- Adds a separate spawn point for the player so that the player reappears in the garden each night rather than where they started the game.
- Adds a debug key for starting the midnight transition, currently defaults to "M".
- Makes the global shaders disable preview when the game stops running.


Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -74,6 +74,13 @@ enum HappinessState
 		Happy,
 	}
 
+	enum BuddyState
+	{
+		Normal,
+		Dead,
+		Adult,
```
brickman1444, line 81: Consider adding an invalid state as the first enum entry. An uninitialized enum defaults to the first value.

Assets/Scripts/Managers/SoundManager.cs
```diff
@@ -94,7 +94,7 @@ AudioSource GetSource()
 
 		foreach ( AudioSource audioSource in audioObjs )
 		{
-			if ( !audioSource.isPlaying )
+			if ( audioSource && !audioSource.isPlaying )
```
brickman1444: A null reference should probably be removed from the list altogether.

brickman1444:
Done reviewing.

Assets/Scripts/Buddy/BuddyStats.cs
randomPoison, line 81: I don't think that's needed in this case, I explicitly initialize the buddies to `Normal`.

Assets/Scripts/Managers/SoundManager.cs
randomPoison: Done.

PR #110, opened 2015-05-02
Background Songs Audio Zones

brickman1444:
Triggers the playing of different background songs upon entering audio zones.

Scene changes required:
- Delete the Garden Music object from the scene hierarchy. It is below Surrounding Lands.
- Add the Audio Zones prefab to the hierarchy. Parent it under whatever but make sure the position and rotation are zeroed out.


Assets/Scripts/Managers/BackgroundMusicManager.cs
```diff
@@ -0,0 +1,81 @@
+﻿using UnityEngine;
+using System.Collections;
+
+public class BackgroundMusicManager : SingletonBehaviour<BackgroundMusicManager>
+{
+	[Tooltip( "Seconds it takes to fade out a song" )]
+	[SerializeField] float fadeOutTime = 0f;
+	[SerializeField] float fadeInTime = 0f;
```
randomPoison, line 9: Is there a reason we'd want to have a different amount of time for fading a song in and out? If one's fading out and another is fading in shouldn't that happen in the same amount of time?

Also, if one has a tooltip the other should too.

Assets/Scripts/Managers/BackgroundMusicManager.cs
```diff
@@ -0,0 +1,81 @@
...
+
+	[ReadOnly]
+	[SerializeField] AudioSource _currentAudioSource = null;
+
+	Coroutine _fadeOutCoroutine = null;
+	Coroutine _fadeInCoroutine = null;
+
+	// Use this for initialization
+	void Start () {
```
randomPoison: Remove unused method.

randomPoison:
I'm done looking this over.

Assets/Scripts/Managers/BackgroundMusicManager.cs
brickman1444, line 9: It's nice to have the control. I currently have the fade out at 5 seconds and the fade in at 4 seconds and it sounds pretty nice. Worst case scenario you adjust two numbers instead of one. Added the second tooltip.

Assets/Scripts/Managers/BackgroundMusicManager.cs
brickman1444: done

PR #111, opened 2015-05-04
Physics

randomPoison:
- Makes the lifter a separate rigid body to better take advantage of Unity's physics constraints.


Assets/Scripts/Actors/ActorPhysics.cs
```diff
@@ -338,7 +347,10 @@ public void DoJump()
 	 */
 	public void AirMovement( Vector3 inputVec, bool forceUp = false )
 	{
+		_rigidbody.useGravity = true;
+
 		FollowBumper();
+		//FollowLifter();
```
brickman1444: Should this be here?

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -111,6 +116,7 @@ public override void Enter()
 
 		public override void Update()
 		{
+			player.animator.SetBool( "isInAir", true );
```
brickman1444: Why is this here instead of the enter method?

brickman1444:
done reviewing

Assets/Scripts/Actors/ActorPhysics.cs
randomPoison: Nope, done.

Assets/Scripts/Player/PlayerControls.cs
randomPoison: I'll change this to an invoke.

PR #113, opened 2015-05-04
Ui stuff

cjacobwade:


Assets/Scripts/Managers/DayCycleManager.cs
```diff
@@ -84,6 +84,12 @@ void MorningReset()
 		Invoke( "StartMidnightOverlay", timeUntilBlackout );
 	}
 
+	public float GetNormalizeCurrentTime()
```
randomPoison: This should be `GetNormalizedCurrentTime()` (you forgot the 'd' in "Normalized"). Also, you should add a comment noting that "normaized" in this case means that it factors out time before the start of the day.

Assets/Scripts/UI/InventoryScrollBar.cs
```diff
@@ -38,7 +38,15 @@ public void NullInventoryBar()
 	void SetIcon( Image image, Sprite icon = null )
 	{
 		image.sprite = icon;
-		image.color = Color.white;
+
+		if( !icon )
+		{
+			image.color = Color.white - (Color)new Vector4( 0f, 0f, 0f, 1f );
```
randomPoison: If you just want to change the alpha of a color we have a `SetAlpha()` extension method for `Color` that would make your intent here more clear.

Assets/Scripts/UI/SkillBar.cs
```diff
@@ -0,0 +1,28 @@
...
+
+	RectTransform _uiBarRectFront = null;
+	RectTransform _uiBarRectBack = null;
+
+	PlayerStats _playerStats = null;
+
+	void Awake()
+	{
+		_uiBarRectBack = GetComponent<RectTransform>();
+		_uiBarRectFront = transform.GetChild(0).GetComponent<RectTransform>();
```
randomPoison: You're using a member/property named `transform` here but I don't see where it's coming from. Is this one of the old deprecated automatic properties?

randomPoison:
Assuming everything works correctly this looks good to me.

Assets/Scripts/Managers/DayCycleManager.cs
cjacobwade: Done

Assets/Scripts/UI/InventoryScrollBar.cs
cjacobwade: Done

Assets/Scripts/UI/SkillBar.cs
cjacobwade: Fixed to use GetComponent<Transform>

PR #114, opened 2015-05-05
Buddy animation logic

brickman1444:
Adds buddy actions to the behavior tree and expands the buddy animation controller. It's not totally perfect yet but it's good for now.

### DIFF
--- a/Assets/Scripts/Nodes/IsDestinationWithinLimits.cs
+++ b/Assets/Scripts/Nodes/IsDestinationWithinLimits.cs
@@ -5,16 +5,16 @@ using BehaviorTree;
 
 public class IsDestinationWithinLimits : LeafNode
 {
-	Vector3 _destination;
+	private Hashtable _data;
 
 	public override void InitSelf( Hashtable data )
 	{
-		_destination = (Vector3)data["destination"];
+		_data = data;
 	}
 
 	public override NodeStatus TickSelf()
 	{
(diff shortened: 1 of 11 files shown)


Assets/Scripts/Nodes/IsDestinationWithinLimits.cs
```diff
@@ -5,16 +5,16 @@
...
 	public override void InitSelf( Hashtable data )
 	{
-		_destination = (Vector3)data["destination"];
+		_data = data;
 	}
 
 	public override NodeStatus TickSelf()
 	{
-		if ( MathUtils.IsWithinInfiniteVerticalCylinders( _destination, LimitsManager.colliders ) )
+		if ( MathUtils.IsWithinInfiniteVerticalCylinders( (Vector3) _data["destination"], LimitsManager.colliders ) )
```
cjacobwade, line 17: I assume this is the fix for the cylinder error that was coming up? What was the issue?

brickman1444, line 17: The destination value was being set in the tick method of another node. So when they were run in sequence it would go Init, Init, Tick, Tick and this node would try to read it in the Init before it was set.

cjacobwade, line 17: Gotcha

PR #132, opened 2015-05-10
Buddy interaction gameplay changes

brickman1444:
Changes how stats are calculated. Stats now have a minimum and the structure of when stats are calculated changed. New emotes are now implemented in game and used to show current hunger level. Emotes are also tinted red to green based on hunger level. Item Faker was also implemented.


randomPoison:
This looks good to me.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -283,56 +263,44 @@ void RestartEmoteRoutine()
...
+	{
+		float t;
+
+		if ( _resources < _idealResourcesRange.max )
+		{
+			t = (float) _resources / _idealResourcesRange.max;
+		}
+		else
+		{
+			t = 1f - (float) ( _resources - _idealResourcesRange.max ) / ( _idealResourcesRange.max - _idealResourcesRange.min );
```
randomPoison, line 291: Rather than doing `_idealResourcesRange.max - _idealResourcesRange.min` you should be able to do `_idealResourceRange.range`. If that's not already a property you should add one.

Assets/Scripts/Buddy/BuddyStats.cs
```diff
@@ -283,56 +263,44 @@ void RestartEmoteRoutine()
...
-			}
+			UpdateHungerEmoteTexture();
+
+			EmoteHunger( _hungerEmoteMaterial );
+		}
+	}
+
+	private void UpdateHungerEmoteTexture()
+	{
+		float t;
```
cjacobwade, line 283: Can you pick a better name for this? Not really sure what's going on here

PR #133, opened 2015-05-11
Tutorial Improvements

randomPoison:
- Adds support for multiple text volumes triggering a single text overlay.
- Adds tutorial text for when a buddy is ready to become and adult.
- Adds tutorial text for the lifters.
- Adds tutorial text for when a buddy is brought to the start area and becomes an adult.
- Fixes a bug that would cause a buddy's adult particles to not play unless the buddy was on the player's back the night before.


Assets/Scripts/Resources/ResourceSpawner.cs
```diff
@@ -64,16 +64,19 @@ void Start()
 
 	void OnDrawGizmos()
 	{
-		Transform trans = GetComponent<Transform>();
-		Gizmos.DrawIcon( trans.position + Vector3.up * 0.5f, "ResourceSpawner.png", true );
+		Transform transform = GetComponent<Transform>();
+
+		Gizmos.matrix = transform.localToWorldMatrix;
+
+		Gizmos.DrawIcon( transform.position + Vector3.up * 0.5f, "ResourceSpawner.png", true );
```
cjacobwade, line 71: Is this set to a weird position because Gizmos.matrix is set to transform.localToWorldMatrix

cjacobwade:
All done

Assets/Scripts/Resources/ResourceSpawner.cs
randomPoison, line 71: So the weird thing I was noticing was that `Gizmos.matrix` wasn't affecting `DrawIcon()`, so the cube was appearing relative to the object, but all of the text icons were showing up at the origin (or `Vector3.up \* 0.5f). Adding the transform's position back in fixed it, but I found it weird.

PR #134, opened 2015-05-12
Buddy interaction gameplay

brickman1444:
This was intended to be balance stuff but I haven't gotten there yet. Balance stuff should just be numbers and will be easy to merge in later. Here is some remaining gameplay changes that I wanted to make sure got in.

Buddy pick up and put down is now on right click. They are not fed when picked up. They are not made happier yet. 

The high point of the buddy's satisfaction with its hunger level is now the midpoint of the ideal range. I think this will send a clearer message to the player.

**Scene changes needed**: If time allows, keep the bushes at the bottom of the sky spire at 1 health. Set the bushes partway up the spire to 2 health. And set the bushes near the top of the spire to 3 health.


randomPoison:
This looks good to me.

PR #136, opened 2015-05-12
Small Playability Tweaks

brickman1444:
Starts to tweak numbers of buddy interactions. I need to do/see more playtesting to tweak more. Adds the full range of item fakers, one for each resource and one for each type of egg. 1-3 on the keyboard is skill fakers. 4-6 is resources. 7-9 is eggs. Added colored egg icons for readability while playing. They look a bit out of place but it's better than before.


randomPoison:
This looks fine. For the record non-code changes don't need to be reviewed, they can just be merged directly (unless you specifically want someone to pull down your changes and test them out).

PR #147, opened 2015-05-21
Showbuild may 19

brickman1444:


Assets/Scripts/Player/PlayerStats.cs
```diff
@@ -24,6 +24,10 @@ public class StatObject
 
 	[ReadOnly]
 	public float rechargeTimer = 0.0f;
+
+	public delegate void StatUpdateCallback( float newValue );
+	public StatUpdateCallback UpdateMaxStat = delegate {};
+	public StatUpdateCallback UpdateStat = delegate {};
```
brickman1444: @cjacobwade i think these should be renamed to something that makes it clear that these are callbacks. Otherwise when you call UpdateStat() it looks like it updates the stat when really it updates the UI.

PR #149, opened 2015-05-24
Animation Improvement

randomPoison:
I hooked up the remaining animations that were not yet in the game.


Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -243,7 +245,12 @@ public override void Update()
 			     player.stats.CanUseStat( Stat.Stamina ) &&
 			     player.physics.ClimbCheck() )
 			{
-				player.physics.ClimbSurface( player.controls.GetMoveInput() );
+				Vector3 moveInput = player.controls.GetMoveInput();
+				player.physics.ClimbSurface( moveInput );
+				player.animator.SetFloat( "moveSpeed", moveInput.magnitude );
+				player.animator.SetFloat( "verticalSpeed", moveInput.z );
+				player.animator.SetFloat( "horizontalSpeed", moveInput.x );
+				player.animator.SetFloat( "slope", Mathf.Abs( Vector3.Dot( moveInput, Vector3.forward ) ) );
```
brickman1444, line 253: What do you use this value for?

Assets/Scripts/Player/PlayerControls.cs
```diff
@@ -243,7 +245,12 @@ public override void Update()
 			     player.stats.CanUseStat( Stat.Stamina ) &&
 			     player.physics.ClimbCheck() )
 			{
-				player.physics.ClimbSurface( player.controls.GetMoveInput() );
+				Vector3 moveInput = player.controls.GetMoveInput();
+				player.physics.ClimbSurface( moveInput );
+				player.animator.SetFloat( "moveSpeed", moveInput.magnitude );
```
brickman1444, line 250: Is this correct? Don't we normally set this to the normalized move speed?

Assets/Scripts/Player/PlayerControls.cs
brickman1444, line 253: Also watch out because you might want transform.forward instead of vector3.forward.

brickman1444:
Done looking this over

Assets/Scripts/Player/PlayerControls.cs
randomPoison, line 253: I use it to blend between horizontal and vertical climb animations. I'm pretty sure `Vector3.forward` is correct in this case because the result from `GetMoveInput()` is a formatted as `<horizontal input, 0.0f, vertical input>` and I want to get a value that goes between 0 and 1 depending on whether the joystick is left/right or up/down.

Assets/Scripts/Player/PlayerControls.cs
randomPoison, line 250: `PlayerPhysics.normalizedMoveSpeed` isn't correct when climbing because it zeroes-out the Y component of the movement vector (essentially getting the horizontal move speed). In this case I want a value between 0 and 1, which using the input gets me.

brickman1444, line 250: ok

Assets/Scripts/Player/PlayerControls.cs
brickman1444, line 253: ok

PR #154, opened 2015-05-30
Ending Cinematics and Adult Buddy Stuff

brickman1444:
- Hooks up the adult volume and the adult manager so that they can properly end the game and trigger the cinematics.
- Creates the cinematic scenes and sets them up to transfer back to the main menu after pressing spacebar.
- Adds a conditional text volume based on whether or not the player has an adult buddy on their back.


Assets/Scripts/Actors/Buddies/CinematicBuddyAnimationTrigger.cs
```diff
@@ -0,0 +1,26 @@
...
+
+	void Start () 
+	{
+		string triggerName = _triggerNames[ triggerIndex ];
+
+		Animator animator = GetComponent<Animator>();
+		animator.Play( triggerName );
+
+		// Make it kind of random while also encouraging it to move through the list and not hit the same thing twice
+		triggerIndex += Random.Range( 1, 3 );
```
randomPoison: If you want to do this in random order it'd probably be better to just shuffle the list and play it in random order.

Assets/Scripts/Actors/Player/PlayerCamera.cs
```diff
@@ -204,7 +204,7 @@ void MoveCameraBehindPlayer( Vector3 actorToCam )
 
 		if( float.IsNaN( turnSpeedMod ) )
 		{
-			Debug.Log( "BAD NUMBER" );
+			Debug.Log( "Avoiding Camera Bug" );
```
randomPoison: This should probably `Debug.LogWarning()`. Also the message should be more descriptive: "PlayerCamera.turnSpeedMod is NaN, setting it to 1".

randomPoison:
I'm done looking this over.

Assets/Scripts/Actors/Buddies/CinematicBuddyAnimationTrigger.cs
brickman1444: Done

Assets/Scripts/Actors/Player/PlayerCamera.cs
brickman1444: Done

PR #157, opened 2015-05-30
Menu screen

randomPoison:
- Adds a MainMenu scene for starting and ending the game.
- Main menu contains "Play" and "Quit" buttons and has a camera flythrough of the level in the background.


brickman1444:
Looks good. The only thing you should add is pull in my changes from the cinematics, then update the LevelUtils list of scene files with the correct name for the main menu scene file.

PR #165, opened 2015-06-03
Lifter Fixes

brickman1444:
Start the lifter at the top and split the entrance and exit volumes so it feels better. Activate a lifter when it is first stepped on at the top.


randomPoison:
:+1:

PR #166, opened 2015-06-03
Buddy Death

brickman1444:
- Update animation controller to properly transition to death state 
- Make dead buddies pushable


randomPoison:
:+1:

PR #170, opened 2015-06-04
Found an optimization trick to save a good amount of render time! Imp…

cjacobwade:
…roving LODs a bit more. Adding in Wyatt's new animations. Adding in parachute with scaling. Fixing midnight overlay bug.


cjacobwade:
Tasha needs the scene so I'm going to go ahead and merge this in

PR #173, opened 2015-06-06
Main Menu Polish

brickman1444:
- Make the main menu more responsive
- Add a button to go to the credits scene to the main menu
- Remove the debug keycode for loading the credits scene in game
- Add enter as a submit button on the main menu
- Add a custom title to the game when built
- Add a custom icon to the game when built
- Make the credits move at a constant speed relative to time
  - Fix the cutting visual which was the only other thing using a Translator
- Load the main menu after the credits are over
- Get rid of the custom input module. Use a standard input module and add a separate component to wrap the input names with the platform. This makes mouse input on the main menu work
- Make escape on the credits scene load the main menu early
- Make submit or cancel buttons on the ending scenes load the credits scene
- Clean up the input array


cjacobwade:
looks good

PR #176, opened 2015-06-09
Buddy Bugs

brickman1444:
- Fix various interactions between aging up, dying and being put down
- Fix adult particles to simulate in world space
- Make the back buddy do proper adult particles instead of cry
- Fix bug where buddies could be spawned all over the world in weird places
- Freeze the buddies face on death
- Get rid of orange buddy color
- Don't highlight dead buddies
- Stop the player from sliding while putting down or picking up buddies
- Scale the buddy up or down while putting down or picking up buddies


randomPoison:
This looks good to me.

PR #177, opened 2015-06-10
Ending and Buddy Fixes

brickman1444:
- Fix the lighting in the bad ending
- Add subtitles to the endings
- Add dead buddies to the bad buddies count
- Change the design of the ending calculation to be first to X whether that's good or bad buddies
- Make very hungry adult buddies count toward bad buddies
- Trigger particles on the back buddy to signal whether the adult will be good or bad
- Play heart particles when picking up a non-adult buddy
- Adjust scaling for pickup and put down to not scale all the way small
- Fix the logic for deciding what buddy sound to play
- Tweak buddy sounds, snore and stomach rumble
- Change buddy animations to be decided by hunger not happiness


randomPoison:
This looks good to me.

